PR #5, opened 2020-05-30
splunk.enterprise_security -> splunk.es

gundalow:


gundalow:
@abadger 1) do I need to `update acd-2.10.0.deps` and `acd-2.10.build` as well?
2) is `2.10/acd.in` still used by anything, given the above two new files?

abadger:
Just update acd.in and I'll regenerate the other two.  I think once we get the hang of this, we won't generate the other two until we are rest to cut the first prerelease.  But i needed to test the docs build do i cut them early.  We can keep regenerating the files until we're ready to freeze, though.

PR #9, opened 2020-06-29
hetzner.hcloud has reached 0.1.0

gundalow:
hetzner.hcloud have done their first release, so we can include that in the next build of `ansible`

https://galaxy.ansible.com/hetzner/hcloud

FYI @LKaemmerling


abadger:
Thanks @LKaemmerling ! Thanks @gundalow !

PR #17, opened 2020-07-10
Enable community.mysql, add infinidat.infinibox

felixfontein:
community.mysql has been released (0.1.0 currently), and some modules were moved from community.general to infinidat.infinibox (ansible_builtin_runtime.yml in both devel and stable-2.10 is updated).


abadger:
Merged, thanks!

PR #18, opened 2020-07-18
Ensure the windows collections include any 1.x.y release

jborean93:
When Ansible 2.10 comes out I am planning on making a 1.0.0 release of both Windows collections. Right now there is `0.2.0` in Galaxy but that's just a way for people to test things out and pick up any critical issues that may have been missed.

### DIFF
--- a/2.10/acd-2.10.build
+++ b/2.10/acd-2.10.build
@@ -3,7 +3,7 @@ _ansible_base_version: 2.10.0b1
 amazon.aws: >=1.0.0,<2.0.0
 ansible.netcommon: >=1.0.0,<2.0.0
 ansible.posix: >=1.0.0,<2.0.0
-ansible.windows: >=0.0.0,<1.0.0
+ansible.windows: >=0.0.0,<2.0.0
 arista.eos: >=1.0.0,<2.0.0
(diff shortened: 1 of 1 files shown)


abadger:
This git repo is largely to record information about what versions of what collections go into each ansible release.  The .deps file is created manually for a release series (2.10.x) and then the .build file and the .deps files are generated automatically from that.

The best way to make sure that the 1.0.0 release of the windows collections make it into ansible-2.10 is to get them released before the beta freeze (tentatively scheduled for 2020-08-18 (Tuesday): Ansible 2.10 Beta freeze. ).  Failing that, if you can say that the 1.0.0 release is backwards compatible with the release which does make it into ansible-2.10 b1, then we could take a PR like this to manually update the .builds file.

If the 1.0.0 releases of the widows collections come out after ansible-2.10.0 is released, then we'd need to have a discussion at the IRC community meeting to see what the community thinks about including it.... (My guess is that if it's backwards compatible with what is in ansible-2.10.0, it would probably be allowed.  If it is not, it probably would not be allowed).

Does that all make sense?  f so, go ahead and close this PR.

jborean93:
Thanks for letting me know, I wasn't aware it was generated automatically.

PR #22, opened 2020-07-26
Extend changelog (and later porting guide) with manual per-release data

felixfontein:
This contains among others all the non-base porting guide entries; see ansible/ansible#70891.


felixfontein:
After rebasing, this PR now shows the differences to the changelog without the manual extra data.

felixfontein:
This PR now only depends on ansible-community/antsibull#158 (and vice versa), so it can be merged without a change to ansible/ansible's stable-2.10.

PR #27, opened 2020-08-03
dellemc.os6 and dellemc.09 are live

gundalow:
They've just been published, so include them

https://galaxy.ansible.com/dellemc
https://github.com/ansible-collections/dellemc.os10/issues/20#issuecomment-667882220


abadger:
Nice.  note that the build file doesn't need to be updated unless we're at the stage where we're going to update versions manually (once beta starts).

PR #33, opened 2020-10-13
Remove bad part of changelog

felixfontein:
Will add a fix to antsibull later which prevents this in the first place.


abadger:
Thanks!

PR #34, opened 2020-11-04
Update changelog

felixfontein:
Updated changelog with the code from ansible-community/antsibull#209

### DIFF
--- a/2.10/CHANGELOG-v2.10.rst
+++ b/2.10/CHANGELOG-v2.10.rst
@@ -1949,11 +1949,13 @@ cisco.meraki
 - meraki - Add optional debugging for is_update_required() method.
 - meraki_admin - Update endpoints for API v1
 - meraki_alert - Manage network wide alert settings.
+- meraki_device - Added deprecation notices to some parameters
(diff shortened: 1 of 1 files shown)


2.10/CHANGELOG-v2.10.rst
```diff
@@ -5217,7 +5222,6 @@ Mongodb
 - community.mongodb.mongodb_shard - Add or remove shards from a MongoDB Cluster
 - community.mongodb.mongodb_shutdown - Cleans up all database resources and then terminates the mongod/mongos process.
 - community.mongodb.mongodb_status - Validates the status of the cluster.
-- community.mongodb.mongodb_stepdown - Step down the MongoDB node from a PRIMARY state.
```
felixfontein, line 5220: Note that this removal is correct, since the module is already listed a bit up. It is not in this specific list as its namespace is correct, as opposed as all the other modules, which end up in the`Community` and then `Mongodb` subsections.

Once ansible-collections/community.mongodb#235 has been fixed, the changelog should look a lot better again.

PR #39, opened 2020-12-01
Add community.okd, a collection which was broken out of the community.general collection, for 2.10.4

abadger:
These are the changes needed to add a new collection to a stable release.  In this PR we add community.okd which was spun out of community.general and thus is allowed into 2.10.x according to the policies approved at the community irc meeting.

I had to do this manually (add the collection name to ansible.in and look on galaxy.ansible.com for the current version and add the collection with the appropriate version range to the ansible-2.10.build file.)  I've opened a ticket on the antsibull repo to add code to do that at some point in the future.


felixfontein:
@abadger does the version range also needs to be added for the other collections we added to ansible.in last(?) week?

abadger:
Yes, it does.  I'm opening a PR for that now.

PR #45, opened 2020-12-23
Add community.google to 2.10 (migrated from community.general)

dmsimard:

PR #47, opened 2021-01-01
Add community.hashi_vault to 2.10 and 3; add community.google to 3

felixfontein:
[community.hashi_vault 1.0.0](https://galaxy.ansible.com/community/hashi_vault) has been released. It contains the hashi_vault lookup plugin that's migrated from community.general.

Also, since 3/ansible.in is already there, new collections should be added there as well. Since community.google wasn't added there in #45, I'm adding it in this PR as well.

CC @briantist


briantist (approved):

gundalow (approved):
These collections contain plugins that were previously in `community.general`, therefor they are allowed in without formal review against [inclusion checklist](https://github.com/ansible-collections/overview/blob/main/collection_requirements.rst)

PR #49, opened 2021-01-09
Prepare changelog for 3.0.0 so that the ancestor is 2.10.0

felixfontein:
Data used for ansible-community/antsibull#231


3/ancestor.deps
```diff
@@ -0,0 +1 @@
+../2.10/ansible-2.10.0.deps
```
felixfontein, line 1: The name of this file is random. Antsibull goes through all `.deps` files and reads them all; and identifies this dependency file as the ancestor by comparing its version with the ancestor version from `changelog.yaml`. Naming the symlink this way should make it more obvious though.

abadger:
This seems good.  Could you also open a PR on the antsibull repo to add it to the docs/build-ansible.rst file?

felixfontein:
@abadger thanks for reviewing and merging!

Yes, I'll do that, including changing the 2.11 there (I think it has 2.11 in it) to 3 ;)

PR #58, opened 2021-01-28
Include dellemc.openmanage collection for Ansible 3.0

dmsimard:
The history of the application as well as the review is available at:
https://github.com/ansible-collections/ansible-inclusion/discussions/3

### DIFF
--- a/3/collection-meta.yaml
+++ b/3/collection-meta.yaml
@@ -1,5 +1,8 @@
 ---
 collections:
+  dellemc.openmanage:
+    maintainers:
+      - rajeevarakkal
   community.sops:
     maintainers:
       - endorama
--- a/3/collection-meta.yaml
+++ b/3/collection-meta.yaml
@@ -1,6 +1,7 @@
 ---
 collections:
   dellemc.openmanage:
+    changelog-url: https://github.com/dell/dellemc-openmanage-ansible-modules/blob/collections/CHANGELOG.rst
     maintainers:
       - rajeevarakkal
   community.sops:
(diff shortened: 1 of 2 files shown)


3/collection-meta.yaml
```diff
@@ -1,5 +1,8 @@
 ---
 collections:
+  dellemc.openmanage:
+    maintainers:
+      - rajeevarakkal
```
felixfontein, line 5:
```suggestion
      - rajeevarakkal
    changelog-url: https://github.com/dell/dellemc-openmanage-ansible-modules/blob/collections/CHANGELOG.rst
```

3/collection-meta.yaml
```diff
@@ -1,5 +1,8 @@
 ---
 collections:
+  dellemc.openmanage:
+    maintainers:
```
felixfontein, line 4:
```suggestion
    changelog-url: https://github.com/dell/dellemc-openmanage-ansible-modules/blob/collections/CHANGELOG.rst
    maintainers:
```

dmsimard: @felixfontein I might have assumed that this wasn't required for collections that are using antsibull-changelog for relnotes given that the other changelogs in the file are markdown. Should we add it anyway ?

felixfontein: They do not use antsibull-changelog, since they do not have a changelogs/changelog.yaml file.

dmsimard: OH, you're right, I don't know how I missed that.

felixfontein: The CHANGELOG.rst file really looks like is created by antsibull-changelog :)

felixfontein (approved):

abadger:
Great!  Thank you!

rajeevarakkal:
@felixfontein , @abadger - Thanks a lot :)

PR #59, opened 2021-01-28
Include kubernetes.core in ansible-3

jillr:
The community.kubernetes will be renamed to kubernetes.core with its future 2.0 release.
The contents of the community.kubernetes repo are being published to both collection
names on Galaxy today (1.x.x series).
community.okd versions >= 1.0.1 depend on the name kubernetes.core.
https://github.com/ansible-collections/community.kubernetes/issues/221

### DIFF
--- a/3/collection-meta.yaml
+++ b/3/collection-meta.yaml
@@ -15,6 +15,7 @@ collections:
       - rndmh3ro
       - schurzi
   kubernetes.core:
-    changelog-url: https://github.com/ansible-collections/community.kubernetes/blob/main/CHANGELOG.rst
     maintainers:
-      - Ansible Cloud Team
+      - jillr
+      - akasurde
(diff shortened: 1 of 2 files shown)


3/collection-meta.yaml
```diff
@@ -14,3 +14,7 @@ collections:
     maintainers:
       - rndmh3ro
       - schurzi
+  kubernetes.core:
+    changelog-url: https://github.com/ansible-collections/community.kubernetes/blob/main/CHANGELOG.rst
+    maintainers:
+      - Ansible Cloud Team
```
abadger, line 20: This needs to be github ids.  The idea is that we need at least one point of contact that we can get a hold of in case there's something wrong with the collection.

abadger:
Looks good!  Merged.

PR #81, opened 2021-11-02
Ansible 5.0.0a3: Dependencies, changelog and porting guide

dmsimard:


dmsimard:
@felixfontein the changelog and porting guide updates for c.g are substantial because of the jump from 3.8.0 to 4.0.0. Could you double check to validate that it's OK ?

felixfontein (approved):

PR #84, opened 2021-11-16
Ansible 5.0.0b2: Dependencies, changelog and porting guide

dmsimard:


dmsimard:
Let's use https://github.com/ansible-community/ansible-build-data/pull/85 instead

PR #91, opened 2021-12-02
Ansible 5.0.1: Dependencies, changelog and porting guide

dmsimard:
We're also including a galaxy-requirements.yaml file for convenience
now.


felixfontein:
I don't think it is a good idea to have that galaxy requirement file unversioned, I think we should have one per Ansible release. But that's something we should probably discuss in community-topics.

dmsimard:
> I don't think it is a good idea to have that galaxy requirement file unversioned, I think we should have one per Ansible release. But that's something we should probably discuss in community-topics.

Thanks for the feedback. I thought about that too but the file is in a git repo which itself is versioned with commits and tags.

I went with that approach so the link doesn't change every release and to make it easier to refer/link to and programmatically use (if need be) since the intent is to provide the collections that would have otherwise been installed by installing the ansible package which is whatever the latest version was.

Hopefully that makes sense.

felixfontein:
Ok, I didn't knew / noticed we had tags in this repo :)

PR #94, opened 2021-12-15
github ci: test builds for both ansible 5 and 6

dmsimard:


dmsimard:
Ah, it's failing because we've changed ``_ansible_base`` to ``_ansible_core`` in the ansible.in/deps file. The role currently checks for ``_ansible_base`` so it would need to support both unless we change the string across every file.

I sent a change to antsibull so we can check for both: https://github.com/ansible-community/antsibull/pull/362

felixfontein:
With that PR merged, tests pass :)

felixfontein:
@dmsimard thanks for improving the tests!

dmsimard:
@felixfontein thanks for reviewing both PRs :)

PR #98, opened 2022-01-12
Ansible 5.2.0: Dependencies, changelog and porting guide

dmsimard:


5/ansible-5.2.0.deps
```diff
@@ -0,0 +1,101 @@
+_ansible_version: 5.2.0
+_ansible_base_version: 2.12.0
```
dmsimard, line 2: Needs to be rebuilt after modifying this manually with 2.12.1.

What bug was this again ?

felixfontein (approved):

dmsimard:
The failure for the pre-commit.ci job is unrelated.

Thanks for reviewing @felixfontein !

felixfontein:
Is there a reason that pre-commit.ci is active for this repo?

dmsimard:
> Is there a reason that pre-commit.ci is active for this repo?

@felixfontein I am not the one who enabled it but it was installed in the ansible-community org and it appears to have been applied to every repository...

PR #106, opened 2022-03-23
Remove unnecessary 6/ansible-6.build file

felixfontein:
The .build file is automatically created when new-ansible is called, which is done by the Ansible 6 tests in CI and when building the first alpha versions. Once 6.0.0a1 has been created this file should be checked in, but not before.


dmsimard:
Sure.

felixfontein:
@dmsimard thanks for reviewing and merging!

PR #107, opened 2022-03-23
Remove superseded collections from Ansible 6

dmsimard:
- community.google -> google.cloud
- community.azure -> azure.azcollection
- community.kubernetes -> kubernetes.core
- servicenow.servicenow -> servicenow.itsm

### DIFF
--- a/6/ansible.in
+++ b/6/ansible.in
@@ -22,7 +22,6 @@ cisco.ucs
 cloud.common
 cloudscale_ch.cloud
 community.aws
-community.azure
 community.ciscosmb
 community.crypto
 community.digitalocean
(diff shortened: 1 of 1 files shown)


dmsimard:
WIP, see https://github.com/ansible-community/community-topics/issues/79

felixfontein (changes requested):
I'm -1 on removing anything without a longer deprecation period. Removing it from Ansible 7 is fine if we deprecate it in 6.0.0 (or in 5.x.y).

mariolenz:
@dmsimard Since Ansible 6 is already released, I think it's too late to remove any collections. Therefor, I think this PR should be closed.

PR #120, opened 2022-05-11
Add vmware.vmware_rest to Ansible 6

mariolenz:
Implements ansible-community/community-topics#99

### DIFF
--- a/5/ansible-5.build
+++ b/5/ansible-5.build
@@ -98,5 +98,6 @@ servicenow.servicenow: >=1.0.0,<2.0.0
 splunk.es: >=1.0.0,<2.0.0
 t_systems_mms.icinga_director: >=1.24.0,<2.0.0
 theforeman.foreman: >=2.2.0,<3.0.0
+vmware.vmware_rest: >=2.1.5,<3.0.0
 vyos.vyos: >=2.6.0,<3.0.0
 wti.remote: >=1.0.0,<2.0.0
--- a/5/ansible.in
+++ b/5/ansible.in
@@ -98,5 +98,6 @@ servicenow.servicenow
 splunk.es
 theforeman.foreman
 t_systems_mms.icinga_director
+vmware.vmware_rest
 vyos.vyos
 wti.remote
--- a/5/collection-meta.yaml
+++ b/5/collection-meta.yaml
@@ -78,3 +78,6 @@ collections:
   community.sap:
     maintainers:
       - rainerleber
+  vmware.vmware_rest:
+    maintainers:
+      - goneri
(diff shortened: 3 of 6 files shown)


Andersson007:
@mariolenz i suggest including it in Ansible 5 as we expect more of its minor releases. In this case we need to follow https://github.com/ansible-community/ansible-build-data#the-current-ansible-major-release

Andersson007:
> @mariolenz i suggest including it in Ansible 5 as we expect more of its minor releases. In this case we need to follow https://github.com/ansible-community/ansible-build-data#the-current-ansible-major-release

felixfontein (changes requested):
Please also update `6/collection-meta.yaml`. And probably also add it to `5/`.

felixfontein (approved):

Andersson007:
@mariolenz thanks for the PR!

 > BTW: I think it would be better if we could get the maintainers automatically from the collections instead of documenting them manually.

Would be nice! Any ideas on how can we do it?

Andersson007:
@felixfontein thanks for reviewing!

felixfontein:
> > BTW: I think it would be better if we could get the maintainers automatically from the collections instead of documenting them manually.
> 
> Would be nice! Any ideas on how can we do it?

It would be nice indeed, but I don't see any way to do that. The only place where we have the list of maintainers is in the discussion for the inclusion application.

mariolenz:
> > > BTW: I think it would be better if we could get the maintainers automatically from the collections instead of documenting them manually.
> > 
> > 
> > Would be nice! Any ideas on how can we do it?
> 
> It would be nice indeed, but I don't see any way to do that. The only place where we have the list of maintainers is in the discussion for the inclusion application.

I agree, it's not possible at the moment. But how about a new requirement for collections that are part of the community package? Like a mandatory `MAINTAINERS` (or similar) file with the list of maintainers.

This way, collections could update the list of maintainers themselves. At the moment, they have to inform us somehow or other. If I'm any juge, that's often forgotten so our list of maintainers is probably outdated.

For example, I didn't find a single "Update maintainers for collection XYZ" commit. There are a lot of collections included and I can't imagine that there hasn't been _any_ maintainer quitting or joining a collection over the last years.

felixfontein:
> For example, I didn't find a single "Update maintainers for collection XYZ" commit. There are a lot of collections included and I can't imagine that there hasn't been _any_ maintainer quitting or joining a collection over the last years.

Most collections aren't mentioned in the collection-meta.yaml file anyway. We only started adding all new collections there (with the maintainers listed in the inclusion application) after we decided to collect maintainer information in collection-meta.yaml so we'd have some folks to ping in case of problems. But I agree, that information can easily get out of date.

If we require a mandatory maintainers file, it should probably be in a machine-readable format to make it possible to automate things (like directly ping the main maintainers in case of general sanity check failures).

PR #128, opened 2022-05-26
add vultr.cloud

resmo:
Also see https://github.com/ansible-collections/ansible-inclusion/discussions/44

/cc @Andersson007


felixfontein (changes requested):
Adding negative review to prevent premature merging.

PR #129, opened 2022-05-30
Avoid including the complete condensed changelog of collections added to Ansible to that Ansible release's changelog and porting guide entries.

felixfontein:
For example, when cisco.dnac is added to Ansible, in the changelog and porting guide release for the version that adds that collection (6.0.0b1) the full condensed changelog of cisco.dnac is contained. This is not helpful for users since there are quite some entries and they do not cover anything that was included in Ansible before (since the collection is new). It's also not helpful for users who manually installed the collection before since they usually didn't install its initial release, but a later version. So it's better to not include all that information.

Result of https://github.com/ansible-community/antsibull/pull/426


Andersson007 (approved):

felixfontein:
@Andersson007 thanks for reviewing this one as well!

PR #135, opened 2022-06-09
fixed typo in versioning information

jehof:
Fixed typo

I think Ansible 5.9.0 uses ansible-core 2.12.6 and not 2.12.0

### DIFF
--- a/5/CHANGELOG-v5.rst
+++ b/5/CHANGELOG-v5.rst
@@ -31,7 +31,7 @@ Added Collections
 Ansible-core
 ------------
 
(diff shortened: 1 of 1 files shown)


felixfontein (changes requested):
This file is automatically generated, so changing it manually is not a good idea.

The corresponding 5.9.0 deps file contains the correct version, so this might be a bug in antsibull. I'll take a look later today.

felixfontein:
Superseded by #136. It looks like something went wrong during the release; the changelog included in the Ansible 5.9.0 distribution on PyPi does not have this problem (it also has some other differences that #136 contains).

@jehof thanks for reporting this!

PR #136, opened 2022-06-09
Re-create 5.9.0 changelog

felixfontein:
5/CHANGELOG-v5.rst seems to be broken and doesn't seem to be the one that was created during the 5.9.0 release process. The 5.9.0 artifact on PyPi also contains the version with this PR included.

@Ompragash do you remember whether something happened while building 5.9.0 that could have caused this?

### DIFF
--- a/5/CHANGELOG-v5.rst
+++ b/5/CHANGELOG-v5.rst
@@ -31,10 +31,10 @@ Added Collections
 Ansible-core
 ------------
 
-Ansible 5.9.0 contains Ansible-core version 2.12.0.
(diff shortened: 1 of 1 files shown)


Ompragash:
@felixfontein I think I know what might caused the issue. I remember David saying it's a bug in antsibull. While generating deps file, antsibull-build command fails to update _ansible_base_version in ansible-5.Y.Z.deps to the latest available core (2.12.6) but holds 2.12.0 and I. had to manually modify it before building the tar package. I think the same occurred with Changelog files too

PR #169, opened 2022-10-12
Ansible 6.5.0: Dependencies, changelog and porting guide

rooftopcellist:


oraNod (approved):
LGTM

Ompragash (approved):
LGTM! Thanks for working on this @rooftopcellist

6/ansible-6.5.0.deps
```diff
@@ -0,0 +1,107 @@
+_ansible_version: 6.5.0
+_ansible_core_version: 2.13.5
+amazon.aws: 3.5.0
```
felixfontein, line 3: I'm missing the `_python` entry that would be added by antsibull 0.51.0. I think it's ok to do the current release as-is, but please make sure to use the latest antsibull release for the next Ansible release :)

felixfontein (approved):

felixfontein:
I'm a bit confused over the timing, though. Shouldn't this release have been made a week ago? (And 6.6.0 will then happen in two weeks?)

6/ansible-6.5.0.deps
rooftopcellist: @felixfontein I used the latest version originally, but CI complained about the python3.9 line, so I reverted back to 0.49.0, which was right before this.

rooftopcellist: This change I think: https://github.com/ansible-community/antsibull/pull/448

rooftopcellist:
You are right, this should have gone out a week ago as per the discussion on irc.  This slipped, partially because of miscommunication and partially because of workload, which is unfortunate. 

I'll merge this for now, but let's chat more in IRC about the best way to move forward schedule-wise for 6.x releases and make sure we have the same plan in mind for the schedule for 7.x releases (as there has been some discussion in IRC around moving to a 4 week release cadence to be a little more aligned with the ansible-core release schedule.

6/ansible-6.5.0.deps
felixfontein: Ah, that's helpful to know. There seems to be a bug in there, which for some reason CI didn't complain about so far. I'll try to look at that today or tomorrow. If anyone else is interested, the error is here: https://github.com/ansible-community/ansible-build-data/actions/runs/3236677877/jobs/5302816585#step:10:98

PR #201, opened 2023-02-14
Re-add google.cloud to Ansible 8

mariolenz:
Keep `google.cloud` in Ansible 8 as discussed in ansible-community/community-topics#105 and voted on in ansible-community/community-topics#195.

### DIFF
--- a/7/changelog.yaml
+++ b/7/changelog.yaml
@@ -90,3 +90,10 @@ releases:
 
         `Porting Guide <https://docs.ansible.com/ansible/devel/porting_guides.html>`_'
     release_date: '2023-01-31'
+  7.3.0:
+    changes:
+      deprecated_features:
+      - Since the google.cloud collection seems to be maintained again, we
+        `cancelled the removal process <https://github.com/ansible-collections/overview/blob/main/removal_from_ansible.rst#cancelling-removal-of-an-unmaintained-collection>`__.
+        So contrary to an earlier announcement, this collection is NOT deprecated and will NOT be removed from Ansible 8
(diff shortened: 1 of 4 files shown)


mariolenz:
@felixfontein Good point, but where? We don't have an `UNdeprecated_features` section.

Should I use the `deprecated_features` section and write there that the deprecation has been canceled?

mariolenz:
What do you think about [this](https://github.com/ansible-community/ansible-build-data/pull/201/commits/c8a88bdff782a1df1f9b1d5fd5f68a8406d291c8):

```
deprecated_features:
 - Since the google.cloud collection seems to be maintained again, we
   `cancelled the removal process <https://github.com/ansible-collections/overview/blob/main/removal_from_ansible.rst#cancelling-removal-of-an-unmaintained-collection>`__.
   So contrary to an earlier announcement, this collection is NOT deprecated and will NOT be removed from Ansible 8
   (https://github.com/ansible-community/community-topics/issues/105).
```

felixfontein (approved):
I think this looks good!

(One thing one could discuss is whether to remove / update the earlier deprecation in the Ansible 7 changelog. I don't really like touching changelogs of released versions, though...)

mariolenz:
> One thing one could discuss is whether to remove / update the earlier deprecation in the Ansible 7 changelog. I don't really like touching changelogs of released versions, though...

I don't think it's a good idea to touch changelogs of released versions, either.

felixfontein:
@mariolenz thanks for implementing this!

PR #205, opened 2023-03-26
CI: use `pip` instead of `poetry install`

gotmax23:
antsibull is moving away from poetry, so we instead use pip to install antsibull and friends.

### DIFF
--- a/.github/workflows/antsibull-build.yml
+++ b/.github/workflows/antsibull-build.yml
@@ -69,17 +69,13 @@ jobs:
         working-directory: antsibull
         run: |
           python3 -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


felixfontein (approved):
:tada:

gotmax23:
Oops, I made a mistake here. We also need to install `../antsibull-core`, and I think `antsibull-docs` wasn't needed to begin with?

felixfontein:
antsibull-docs was used in a previous version, it can be removed completely. And yeah, it needs antsibull-core, which right now is installed from PyPI I guess...

gotmax23:
https://github.com/ansible-community/ansible-build-data/pull/206

Issue #229, opened 2023-05-09
ansible 5 build data incorrectly(?) lists _ansible_base_version


dnmvisser:
I'm trying to use the various files here to do a custom install.
Instead of pip installing ansible, I install ansible-core, and then add collections to that.

So for example, for ansible 7.5.0, I would look at 
https://github.com/ansible-community/ansible-build-data/blob/main/7/ansible-7.5.0.deps.
This lists `_ansible_core_version: 2.14.5`, so I would `pip install ansible-core==2.14.5`. This works fine.

But, the deps files here for major version 5 list `ansible-base`, which I think is incorrect, because ansible 5 in reality is using `ansible-core`... The issue is here:

https://github.com/ansible-community/ansible-build-data/blob/main/5/ansible-5.10.0.deps#L2

Trying to install that version will fail as it does not exist:

```
install ansible-base==2.12.7
ERROR: Could not find a version that satisfies the requirement ansible-base==2.12.7 (from versions: 0.0.1a1, 0.0.1a2, 2.10.0b1, 2.10.0rc1, 2.10.0rc2, 2.10.0rc3, 2.10.0rc4, 2.10.0, 2.10.1rc1, 2.10.1rc2, 2.10.1rc3, 2.10.1, 2.10.2rc1, 2.10.2, 2.10.3rc1, 2.10.3, 2.10.4rc1, 2.10.4, 2.10.5rc1, 2.10.5, 2.10.6rc1, 2.10.6, 2.10.7rc1, 2.10.7, 2.10.8rc1, 2.10.8, 2.10.9rc1, 2.10.9, 2.10.10rc1, 2.10.10, 2.10.11rc1, 2.10.11, 2.10.12rc1, 2.10.12, 2.10.13rc1, 2.10.13, 2.10.14rc1, 2.10.14, 2.10.15rc1, 2.10.15, 2.10.16rc1, 2.10.16, 2.10.17rc1, 2.10.17)
ERROR: No matching distribution found for ansible-base==2.12.7
```

I don't know how those `_ansible_base_version` strings ended up there, but they do not seem to be correct....

felixfontein:
This is intentional. The option was only renamed at a later version of antsibull, so `_ansible_base_version` was needed at that point.

felixfontein:
[...]

PR #230, opened 2023-05-09
Use ansible-core instead of ansible-base in v5 deps files

dnmvisser:
This should fix https://github.com/ansible-community/ansible-build-data/issues/229.


felixfontein (changes requested):
The names are correct and should not be changed.

dnmvisser:
OK, i will have to work around it then. thx for clarifying!

Issue #223, opened 2023-04-27
[Tracker] Collection tagging policy violations


gotmax23:
I plan to use this issue as a tracker for collections that need to be pinned to an earlier release due to not tagging releases in their respective Github repositories.

More to come!

gotmax23:
cyberark.pas is *again* violating this guideline (https://github.com/ansible-community/ansible-build-data/pull/235#issuecomment-1556293107).

felixfontein:
dellemc.enterprise_sonic needs to be pinned to 2.0.0 in both Ansible 7 and Ansible 8 for two violations:
- (applies only to Ansible 8) https://github.com/ansible-community/ansible-build-data/issues/223
- (applies to both Ansible 7 and 8) https://github.com/ansible-collections/dellemc.enterprise_sonic/issues/268

felixfontein:
There has been a new release of dellemc.enterprise_sonic (2.2.0) which reverts the breaking change and declares compatibility with `ansible.netcommon` `>5.0.0`. This fixes the problems with Ansible 8 (except that the breaking change will still be mentioned in the next porting guide, even though it is no longer a breaking change - I guess we have to add an Ansible changelog entry for that), but has the downside of not allowing this release for the next Ansible 7 release since that one has ansible.netcommon 4.x.y. So we have to keep the pin for Ansible 7 (though now it has another reason), and undo it (with a changelog entry) for Ansible 8.

felixfontein:
The issue with hetzner.hcloud is resolved (https://github.com/ansible-collections/hetzner.hcloud/issues/217, #267).

gotmax23:
> inspur.ispim 2.0.1 is not tagged in https://github.com/ispim/inspur.ispim

https://github.com/ispim/inspur.ispim/issues/24

felixfontein:
- cisco.intersight 2.0.2 is not tagged in https://github.com/CiscoDevNet/intersight-ansible
- cisco.meraki 2.16.0 is not tagged in https://github.com/CiscoDevNet/ansible-meraki

felixfontein:
cisco.meraki moved to a new repository: https://github.com/meraki/dashboard-api-ansible - I'll create PRs for ansible-build-data to adjust it.

felixfontein:
cisco.intersight was now tagged, #289 removed the ignore for the missing tags.

felixfontein:
- https://github.com/ansible-collections/netapp.aws/issues/8
[...]

PR #240, opened 2023-05-23
Pin dellemc.enterprise_sonic to 2.0.0 in Ansible 7

felixfontein:
Ref: https://github.com/ansible-collections/dellemc.enterprise_sonic/issues/268.
Ref: #223


gotmax23 (approved):
Ack. Thanks!

PR #241, opened 2023-05-23
Pin hetzner.hcloud to 1.11.0 in Ansible 8

felixfontein:
Ref: https://github.com/ansible-collections/hetzner.hcloud/issues/217
Ref: #223


gotmax23 (approved):
Ack. Thanks!

PR #242, opened 2023-05-26
Deprecate netapp.aws

mariolenz:
Deprecate netapp.aws as discussed in ansible-community/community-topics#223 and voted on in ansible-community/community-topics#229

### DIFF
--- a/8/changelog.yaml
+++ b/8/changelog.yaml
@@ -69,3 +69,10 @@ releases:
 
         `Porting Guide <https://docs.ansible.com/ansible/devel/porting_guides.html>`_'
     release_date: '2023-05-23'
+  8.0.0:
+    changes:
+      deprecated_features:
+        - The netapp.aws collection is considered unmaintained and will be removed
+          from Ansible 10 if no one starts maintaining it again before Ansible 10. See
+          `the removal process for details on how this works <https://github.com/ansible-collections/overview/blob/main/removal_from_ansible.rst#cancelling-removal-of-an-unmaintained-collection>`__
+          (https://github.com/ansible-community/community-topics/issues/223).


8/changelog.yaml
```diff
@@ -69,3 +69,10 @@ releases:
 
         `Porting Guide <https://docs.ansible.com/ansible/devel/porting_guides.html>`_'
     release_date: '2023-05-23'
+  8.0.0:
+    changes:
+      deprecated_features:
+        - The netapp.aws collection is considered unmaintained and will be removed
+          from Ansible 10 if no one starts maintaining it again before Ansible 10. See
```
felixfontein, line 76: I think we can already remove it from Ansible 9. (The process description says: "If X.0.0 will be released next, set Y=X+1." - https://github.com/ansible-collections/overview/blob/main/removal_from_ansible.rst#process-3)

felixfontein: Hmm, but then, the vote was explicit for removal from Ansible 10... (https://github.com/ansible-community/community-topics/discussions/229)

mariolenz: I would stay with removing the collection from 10, since this is what the vote was about. BTW: I've also mentioned Ansible 10 in ansible-collections/netapp.aws#7.

If there had been a delay in the Bullhorn issue, we might not have been able to finish the process before the Ansible 9 release :shrug: 

I wanted to be on the safe side, so 10 instead of 9.

felixfontein (approved):

mariolenz:
Thanks for reviewing @felixfontein!

felixfontein:
I created an issue to track this: #243

PR #246, opened 2023-06-10
Deprecate netapp.aws in 7.7.0

mariolenz:
I thought that 7.6.0 would be the last 7.x release, but since it looks like there will be a 7.7.0 we should mention the deprecation of netapp.aws there.

This is kind of a backport of #242


felixfontein (approved):

felixfontein:
@mariolenz thanks!

gotmax23:
Thanks all!

PR #254, opened 2023-06-23
Remove deprecated servicenow.servicenow from Ansible 9

mariolenz:
See ansible-community/community-topics#246 for more information.

### DIFF
--- a/8/ansible-8.build
+++ b/8/ansible-8.build
@@ -99,7 +99,6 @@ purestorage.flasharray: >=1.18.0,<2.0.0
 purestorage.flashblade: >=1.11.0,<2.0.0
 purestorage.fusion: >=1.4.0,<2.0.0
 sensu.sensu_go: >=1.13.0,<2.0.0
-servicenow.servicenow: >=1.0.0,<2.0.0
 splunk.es: >=2.1.0,<3.0.0
 t_systems_mms.icinga_director: >=1.32.0,<2.0.0
 theforeman.foreman: >=3.10.0,<4.0.0
--- a/8/ansible-8.build
+++ b/8/ansible-8.build
@@ -99,6 +99,7 @@ purestorage.flasharray: >=1.18.0,<2.0.0
 purestorage.flashblade: >=1.11.0,<2.0.0
 purestorage.fusion: >=1.4.0,<2.0.0
 sensu.sensu_go: >=1.13.0,<2.0.0
+servicenow.servicenow: >=1.0.0,<2.0.0
 splunk.es: >=2.1.0,<3.0.0
 t_systems_mms.icinga_director: >=1.32.0,<2.0.0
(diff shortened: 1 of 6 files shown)


8/ansible-8.build
```diff
@@ -99,7 +99,6 @@ purestorage.flasharray: >=1.18.0,<2.0.0
 purestorage.flashblade: >=1.11.0,<2.0.0
 purestorage.fusion: >=1.4.0,<2.0.0
 sensu.sensu_go: >=1.13.0,<2.0.0
-servicenow.servicenow: >=1.0.0,<2.0.0
```
felixfontein, line 102: The collection MUST NOT be removed from Ansible 8. This is a breaking change and thus not acceptable since Ansible sticks to semantic versioning.

felixfontein (changes requested):

8/ansible-8.build
mariolenz: I don't think this is the only place we have to change in this case... let me have a look and do another commit.

mariolenz: @felixfontein I think it should be OK now.

8/changelog.yaml
```diff
@@ -95,3 +95,8 @@ releases:
 
         `Porting Guide <https://docs.ansible.com/ansible/devel/porting_guides.html>`_'
     release_date: '2023-06-22'
+  8.2.0:
+    changes:
+      deprecated_features:
+      - The deprecated servicenow.servicenow collection has been removed from Ansible 7, but accidentally re-added to Ansible 8. It will be removed again from Ansible 9.
```
felixfontein, line 101:
```suggestion
      - The deprecated servicenow.servicenow collection has been removed from Ansible 7, but accidentally re-added to Ansible 8. It will be removed again from Ansible 9
```

9/changelog.yaml
```diff
@@ -15,3 +15,5 @@ releases:
         - "``community.skydive`` was considered unmaintained and removed from Ansible 9 as per the
           `removal from Ansible process <https://github.com/ansible-collections/overview/blob/main/removal_from_ansible.rst#unmaintained-collections>`_.
           Users can still install this collection with ``ansible-galaxy collection install community.skydive``."
+        - The deprecated servicenow.servicenow collection has been removed from Ansible 7, but accidentally re-added to Ansible 8. It has been removed again from Ansible 9.
```
felixfontein, line 18:
```suggestion
        - The deprecated servicenow.servicenow collection has been removed from Ansible 7, but accidentally re-added to Ansible 8. It has been removed again from Ansible 9
```

felixfontein (commented):
Two small changes, after that we can merge it.

felixfontein (approved):

felixfontein:
Thanks a lot for fixing this! I really wonder how that collection managed to sneak back in...

PR #260, opened 2023-07-09
9: add changelog entry for setup.cfg configuration

gotmax23:
Relates: https://github.com/ansible-community/antsibull/pull/530

### DIFF
--- a/9/changelog.yaml
+++ b/9/changelog.yaml
@@ -2,6 +2,13 @@ ancestor: 8.0.0
 releases:
   9.0.0a1:
     changes:
+      minor_changes:
+        - Move setuptools configuration into the declarative ``setup.cfg``
+          format. ``ansible`` sdists still contain a stub ``setup.py`` file,
+          but we recommend that users move to tools like pip and build and the
+          PEP 517 interface instead of setuptools' deprecated ``setup.py``
+          interface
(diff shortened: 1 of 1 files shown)


9/changelog.yaml
```diff
@@ -2,6 +2,13 @@ ancestor: 8.0.0
 releases:
   9.0.0a1:
     changes:
+      minor_changes:
```
felixfontein, line 5: I wonder whether this should be moved to
```suggestion
      major_changes:
```
so it shows up in the porting guide? On the other hand, this should not really affect anyone directly. WDYT?

gotmax23: Yeah, it shouldn't really affect anyone except maybe Linux distributions that patch the setup.py in [some way](https://src.fedoraproject.org/rpms/ansible/blob/0321735d7fc291639db0b22572e0d5bd812b7d97/f/ansible.spec#_91). I'm fine with adding it to the Porting Guide, but I don't think it *needs* to be there.

felixfontein: In that case, let's keep it as-is.

felixfontein (approved):

felixfontein:
We should announce this (again) in #ansible-packaging. Besides that, I think it's good to go!

PR #261, opened 2023-07-12
replace t_systems_mms with telekom_mms

rndmh3ro:
We renamed our company and thus the collection-namespace here: https://github.com/ansible/galaxy/issues/3174

The collection now lives here: https://github.com/telekom-mms/ansible-collection-icinga-director

I released a new major version 2.0.0 in the old namespace (https://galaxy.ansible.com/t_systems_mms/icinga_director) where I added the plugin-routing to the new namespace. I also deprecated the collection in the galaxy.

This way I hope in Ansible 8 the old collection will stay the same while in Ansible 9 the new collection will be used.

### DIFF
--- a/9/changelog.yaml
+++ b/9/changelog.yaml
@@ -34,3 +34,9 @@ releases:
           from Ansible 10 if no one starts maintaining it again before Ansible 10. See
           `the removal process for details on how this works <https://github.com/ansible-collections/overview/blob/main/removal_from_ansible.rst#cancelling-removal-of-an-unmaintained-collection>`__
           (https://github.com/ansible-community/community-topics/issues/234).
+        - The collection ``t_systems_mms.icinga_director`` has been renamed to ``telekom_mms.icinga_director``.
+          For now both collections are included in Ansible. The content in ``t_systems_mms.icinga_director``
+          has been replaced with deprecated redirects to the new collection in Ansible 9.0.0,
+          and these redirects will be removed from Ansible 11. Please update your
+          FQCNs for ``t_systems_mms.icinga_director``.
+
--- a/8/ansible-8.build
+++ b/8/ansible-8.build
@@ -102,6 +102,7 @@ sensu.sensu_go: >=1.13.0,<2.0.0
 servicenow.servicenow: >=1.0.0,<2.0.0
 splunk.es: >=2.1.0,<3.0.0
 t_systems_mms.icinga_director: >=1.32.0,<2.0.0
+telekom_mms.icinga_director: >=1.32.0,<2.0.0
 theforeman.foreman: >=3.10.0,<4.0.0
 vmware.vmware_rest: >=2.3.0,<3.0.0
 vultr.cloud: >=1.7.0,<2.0.0
--- a/8/ansible.in
+++ b/8/ansible.in
@@ -100,6 +100,7 @@ servicenow.servicenow
 splunk.es
 theforeman.foreman
 t_systems_mms.icinga_director
+telekom_mms.icinga_director
 vmware.vmware_rest
 vultr.cloud
 vyos.vyos
--- a/8/changelog.yaml
+++ b/8/changelog.yaml
@@ -118,3 +118,9 @@ releases:
         from Ansible 10 if no one starts maintaining it again before Ansible 10. See
         `the removal process for details on how this works <https://github.com/ansible-collections/overview/blob/main/removal_from_ansible.rst#cancelling-removal-of-an-unmaintained-collection>`__
         (https://github.com/ansible-community/community-topics/issues/234).
+      - The collection ``t_systems_mms.icinga_director`` has been renamed to ``telekom_mms.icinga_director``.
+        For now both collections are included in Ansible. The content in ``t_systems_mms.icinga_director``
+        will be replaced with deprecated redirects to the new collection in Ansible 9.0.0,
+        and these redirects will eventually be removed from Ansible. Please update your
+        FQCNs for ``t_systems_mms.icinga_director``.
+
--- a/8/collection-meta.yaml
+++ b/8/collection-meta.yaml
@@ -43,6 +43,12 @@ collections:
       - rndmh3ro
       - schurzi
     repository: https://github.com/T-Systems-MMS/ansible-collection-icinga-director
+  telekom_mms.icinga_director:
+    changelog-url: https://github.com/telekom-mms/ansible-collection-icinga-director/blob/master/CHANGELOG.md
+    maintainers:
+      - rndmh3ro
+      - schurzi
+    repository: https://github.com/telekom-mms/ansible-collection-icinga-director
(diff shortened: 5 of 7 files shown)


rndmh3ro:
Thanks @mariolenz, I didn't know about the docs you linked to.

I hope I've done everything right, now.

mariolenz (approved):
LGTM at the moment, but I'll ask some people if I'm right.

gotmax23:
Our CI is currently failing with:

>  t_systems_mms.icinga_director 2.0.0 is not tagged in https://github.com/T-Systems-MMS/ansible-collection-icinga-director

Please fix the issue so that the collection complies with the [repository management][1] section of the Collection Requirements:

[1]: https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_requirements.html#repository-management

 > Every collection MUST have a public git repository. Releases of the collection MUST be tagged in said repository. This means that releases MUST be `git tag`ed and that the tag name MUST exactly match the Galaxy version number. Tag names MAY have a `v` prefix, but a collection's tag names MUST have a consistent format from release to release.
 >
 > Additionally, collection artifacts released to Galaxy MUST be built from the sources that are tagged in the collection's git repository as that release. Any changes made during the build process MUST be clearly documented so the collection artifact can be reproduced.

mariolenz:
Thanks @gotmax23, I didn't spot that!

@rndmh3ro It looks like you've moved and redirected https://github.com/T-Systems-MMS/ansible-collection-icinga-director to https://github.com/telekom-mms/ansible-collection-icinga-director and this causes some trouble. I think it would have been easier to keep `t_systems_mms.icinga_director` in [T-Systems-MMS](https://github.com/T-Systems-MMS), deprecate it there and tag it with `2.0.0`.

rndmh3ro:
> @rndmh3ro It looks like you've moved and redirected https://github.com/T-Systems-MMS/ansible-collection-icinga-director to https://github.com/telekom-mms/ansible-collection-icinga-director and this causes some trouble. I think it would have been easier to keep t_systems_mms.icinga_director in [T-Systems-MMS](https://github.com/T-Systems-MMS), deprecate it there and tag it with 2.0.0.

But how would I then move it and also publish a telem_mms.icinga_director version 2.0.0?

I don't know how to fix this whole mess. :)

felixfontein:
I've created #264 to ignore the tag error.

The 2.0.0 release seems to have some more problems, for example the deprecation is done **only** in meta/runtime.yml, but not in the modules themselves. This means that the deprecation is completely invisible for Ansible 2.9, and that it will not be visible on the docsite. Also the inventory plugin hasn't been deprecated. There's also no changelog entry for 2.0.0.

9/changelog.yaml
```diff
@@ -24,3 +24,9 @@ releases:
           Users can still install this collection with ``ansible-galaxy collection install community.skydive``."
         - The deprecated servicenow.servicenow collection has been removed from Ansible 7, but accidentally re-added to Ansible 8. It has been removed again from Ansible 9
           (https://github.com/ansible-community/community-topics/issues/246).
+      deprecated_features:
+        - The collection ``t_systems_mms.icinga_director`` has been renamed to ``telekom_mms.icinga_director``.
+            For now both collections are included in Ansible. The content in ``t_systems_mms.icinga_director``
+            will be replaced with deprecated redirects to the new collection in Ansible 9.0.0,
```
felixfontein, line 30:
```suggestion
            have been replaced with deprecated redirects to the new collection in Ansible 9.0.0,
```

9/changelog.yaml
```diff
@@ -24,3 +24,9 @@ releases:
...
+            and these redirects will eventually be removed from Ansible. Please update your
```
felixfontein, line 31: Let's fix a version when this will happen:
```suggestion
            and these redirects will be removed from Ansible 11. Please update your
```

rndmh3ro:
> The 2.0.0 release seems to have some more problems, for example the deprecation is done only in meta/runtime.yml, but not in the modules themselves. This means that the deprecation is completely invisible for Ansible 2.9, and that it will not be visible on the docsite. Also the inventory plugin hasn't been deprecated. There's also no changelog entry for 2.0.0.

Thanks, @felixfontein. I'll try to fix these. Should I release a 2.0.1 in Galaxy for the fixes?

felixfontein:
@rndmh3ro that would probably be better, especially the changelog will be useful for Ansible 9.0.0 since the information will then show up in its combined changelog.

mariolenz:
@rndmh3ro It looks like there are some conflicts that must be resolved. Can you please rebase? I guess this is because of the changes to `changelog.yaml` in #262, #269 and #271.

I'm not 100% sure, but you'll see when you actually do the rebase.

rndmh3ro:
@felixfontein, I now forked the collection from the new namespace to the old one and created the following PR:
https://github.com/T-Systems-MMS/ansible-collection-icinga-director/pull/1

If this PR can be merged, I'll do this. Then I'll release a 2.0.1 in the old collection. Then I'll update the build-data.
Is this the correct way?

mariolenz:
I'm really sorry, but we've just released 8.2.0. Could you please rebase your PR an move the deprecation to [8.3.0](https://github.com/ansible-community/ansible-build-data/blob/8f85dd29cbd62157a15d39e3891314d87bc2ad63/8/changelog.yaml#L114-L120)?

BTW there are two suggested changes, could you please have a look at them?

> I now forked the collection from the new namespace to the old one and created the following PR: [T-Systems-MMS/ansible-collection-icinga-director#1](https://github.com/T-Systems-MMS/ansible-collection-icinga-director/pull/1)
> 
> If this PR can be merged, I'll do this. Then I'll release a 2.0.1 in the old collection. Then I'll update the build-data. Is this the correct way?

@felixfontein What do you think about this, would it help?

rndmh3ro:
I rebased and updated the PR with the suggestions.

rndmh3ro:
And I also released a [2.0.1](https://galaxy.ansible.com/t_systems_mms/icinga_director) of the old collection, hopefully doing everything right.

When running a playbook with the old modules in version 2.0.1, I get:

```
[DEPRECATION WARNING]: t_systems_mms.icinga_director.icinga_serviceset has been deprecated. Use telekom_mms.icinga_director.icinga_serviceset instead. This feature will be removed from
t_systems_mms.icinga_director in version 3.0.0. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.

```

felixfontein (approved):

felixfontein:
CI should be fixed by #273.

felixfontein:
Restart CI.

mariolenz:
@felixfontein Since you've approved and the CI seems to be happy, I've squashed and merged.

Please let me know if you think this has been a mistake. I'll undo in this case.

felixfontein:
No, everything is fine! Thanks everyone for getting this done!

PR #265, opened 2023-07-14
Initial (Github Actions) Workflows for Ansible Community Package release  

anweshadas:
## Introduction 

The Release Manager(s) (RM) for Ansible Community Package, are always the Ansible Community Team members, who are Red Hatters. But with this endeavor we are trying to open up the release process to the community members. So an interested and designated group of people can be Release Ansible.
Who can be the release managers for Ansible Community Package?

Anyone from the Ansible Community can be the RM for Ansible Community Package. We will be forming a Release Manager Working Group from the community who will be not only doing the release but also taking the decisions and responsible for work in and around the release process (barring certain reasonable restrictions and limitations). 

## Automation of the current release process

We are trying to automate the current release process with Github Actions (GHA). We wrote the initial workflow(s). This workflow tries to follow the current release process as much as possible. 

#### Step 1 

Run the [ansible-release workflow] to create the release PR to the ansible-build-data repo.

[ansible-release workflow]:
https://github.com/anweshadas/ansible-build-data/blob/test/.github/workflows/ansible-release.yml

#### Step 2

After the PR is merged by a maintainer, run [upload-to-pypi] to upload the [distribution packages] to PyPI.

[upload-to-pypi]:
https://github.com/anweshadas/ansible-build-data/blob/test/.github/workflows/upload-to-pypi.yml

#### Step 3

After uploading the source to PyPI run [docsbuild-release workflow] to  create PR to ansible-documentation.

[docsbuild-release workflow]:
https://github.com/anweshadas/ansible-documentation/blob/test/.github/workflows/docsbuild-release.yaml

### Plan for betterment this workflow 

Here is the plan to improve the intial workflow

The process needs to be divided in two parts: Pre-Release and Release

Pre- Release includes:

- Creating PR in ansible-build-data
- Merging (manually) the above mentioned PR

Pre-requisite : A new ansible-core release will trigger the Pre-Release workflow

Release workflow includes:

- Make and upload GHA artifact
- Creating PR in ansible-documentation
- Releasing the tar in test.pypi.org
- Releasing the tar to PyPI
- Creating the Git Tag
- Releasing the [distribution packages] to PyPI

Pre-requisite : The release workflow will run (manually) only after the merging of ansible-build-data PR.

[distribution packages]:
https://packaging.python.org/en/latest/glossary/#term-Distribution-Package

The new upstream recommendations for building and publishing from github action to (Test)PyPI:
https://github.com/pypa/packaging.python.org/pull/1261


.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install ansible-core . ../antsibull-core ../antsibull-changelog
```
gotmax23, line 64: The release workflow should install antsibull from PyPI:

```suggestion
          python3 -m pip install --upgrade pip
          python3 -m pip install ansible-core antsibull
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install ansible-core . ../antsibull-core ../antsibull-changelog
```
gotmax23, line 72: See comment above.

```suggestion
          python3 -m pip install --upgrade pip
          python3 -m pip install ansible-core antsibull
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+          ansible-galaxy collection install 'git+https://github.com/ansible-collections/community.general.git'
+
+
+      - name: Create sdist dir using the deps file created in the ansible-build-data Release PR
+        run: |
+          antsibull-build rebuild-single "${{ inputs.ansible_version }}" --data-dir "${BUILD_DATA_DIR}" --build-file "${BUILDFILE}" --deps-file "${DEPSFILE}" --sdist-dir antsibull/sdist --debian
```
gotmax23, line 78: Hmm, I don't like this. Ideally, the previous job would save the artifacts that are generated by the release playbook and then this would simply upload them. As we stand, this is missing `--tags-file` and the `--build-file` and `--deps-file` args are redundant.

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Upload Ansible sdist to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          repository-url: https://test.pypi.org/legacy/
+          packages-dir: antsibull/sdist
```
gotmax23, line 84: What's the idea here? This is not how test PyPI is supposed to be used. Why not have this upload the artifacts to actual PyPI?

.github/workflows/upload-to-pypi.yml
gotmax23: Also, the `--sdist-dir` should probably be `antsibull/dist`. We also build wheels.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+# Run the playbook according to the current release process
+
+      - name: Test building a release with the defaults
+        working-directory: antsibull
+        run: |
+          ansible-playbook -vv playbooks/build-single-release.yaml \
+          -e antsibull_data_reset=false \
+          -e 'antsibull_ansible_version="${{ inputs.ansible_version }}"' \
+          -e 'antsibull_data_dir="{{ antsibull_data_git_dir }}/${{ inputs.ansible_major_version }}"' \
+          -e 'antsibull_build_file="ansible-${{ inputs.ansible_major_version }}.build"'
```
gotmax23, line 83: `antsibull_build_file` should not need to be manually specified.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+          repository: ansible-community/antsibull-core
+          ref: main
+          path: antsibull-core
+
+      - name: Check out antsibull-changelog
+        uses: actions/checkout@v3
+        with:
+          repository: ansible-community/antsibull-changelog
+          ref: main
+          path: antsibull-changelog
```
gotmax23, line 41: These don't need to be checked out if we're just using the playbook from antsibull.

```suggestion
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+          repository: ansible-community/antsibull-core
+          ref: main
+          path: antsibull-core
+
+      - name: Check out antsibull-changelog
+        uses: actions/checkout@v3
+        with:
+          repository: ansible-community/antsibull-changelog
+          ref: main
+          path: antsibull-changelog
```
gotmax23, line 47: See the comment above.

``` suggestion
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Pre-create sdist directory
+        run: |
+          mkdir -p antsibull/build
+          mkdir -p antsibull/sdist
```
gotmax23, line 52:
```suggestion
          mkdir antsibull/build
          mkdir antsibull/sdist
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
+name: ansible-release
+on:
+  workflow_dispatch:
+    inputs:
+      ansible_version:
+        description: 'Release Version. Example : 11.1.0'
+        required: true
+      ansible_major_version:
+        description: 'Exmaple 11'
+        required: true
```
gotmax23, line 10: Future enhancement: make the workflow determine `ansible_version` from `ansible_major_version`,

.github/workflows/ansible-release.yml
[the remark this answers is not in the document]
webknjaz: Global permissions would better be empty and only jobs that need these elevated ones should declare them.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+env:
+  CI_COMMIT_MESSAGE: 'Ansible "${{ inputs.ansible_version }}": Dependencies, changelog and porting guide'
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
webknjaz: LOL I commented the same thing on the sibling PR, I guess many of those comments apply here too.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
+name: ansible-release
+on:
+  workflow_dispatch:
+    inputs:
+      ansible_version:
+        description: 'Release Version. Example : 11.1.0'
+        required: true
+      ansible_major_version:
+        description: 'Exmaple 11'
```
webknjaz, line 9:
```suggestion
        description: Example 11
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
+name: ansible-release
+on:
+  workflow_dispatch:
+    inputs:
+      ansible_version:
+        description: 'Release Version. Example : 11.1.0'
```
webknjaz, line 6:
```suggestion
        description: >-
          Release Version. Example: 11.1.0
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
+name: ansible-release
+on:
+  workflow_dispatch:
+    inputs:
+      ansible_version:
```
webknjaz, line 5:
```suggestion
      ansible-version:
```

.github/workflows/ansible-release.yml
webknjaz: Maybe vice versa?

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+      ansible_major_version:
+        description: 'Exmaple 11'
+        required: true
+env:
+  CI_COMMIT_MESSAGE: 'Ansible "${{ inputs.ansible_version }}": Dependencies, changelog and porting guide'
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  build:
+    name: 'Build Ansible community distribution (${{ inputs.ansible_version }})'
```
webknjaz, line 17: Too long to fit in the UI

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+        run: mkdir -p antsibull/build
+
+# We need to checkout the docs repo here
+
+      # This is where the antsibull build-release role expects it by default
+      - name: Check out ansible-build-data under antsibull build directory
+        uses: actions/checkout@v3
+        with:
+          path: antsibull/build/ansible-build-data
+          ref: main
```
webknjaz, line 53: Why is `ref` needed? Pretty sure this would be the default anyway.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install --upgrade pip
```
webknjaz, line 63: In general, upgrading pip isn't really needed.

.github/workflows/ansible-release.yml
gotmax23: Oops, that's what I meant in the first place.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+          python3 -m pip install ansible-core . ../antsibull-core ../antsibull-changelog
+          ansible-galaxy collection install 'git+https://github.com/ansible-collections/community.general.git'
+
+      - name: Checking out to a new branch & setting the user details
+        run: |
+          cd antsibull/build/ansible-build-data
+          git checkout -b "publish-${{ inputs.ansible_version }}"
+          git config --global user.name "${{github.actor}}"
+          git config --global user.email "${{github.actor}}@users.noreply.github.com"
```
webknjaz, line 72: Same comments as in the other PR:
1) the user
2) the working dir setting

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+        env:
+          # Make result better readable
+          ANSIBLE_CALLBACK_RESULT_FORMAT: yaml
+
+      - name: Commit ansible-build-data and push the changes to github
+        run: |
+         cd "antsibull/build/ansible-build-data/${{ inputs.ansible_major_version }}"
+         git add *.rst *.build *.deps *.yaml *.in validate-tags-ignores
+         git commit -a -m "${{ env.CI_COMMIT_MESSAGE }}"
+         git push origin "publish-${{ inputs.ansible_version }}"
```
webknjaz, line 94: working dir / double quotes

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+
+        env:
+          # Make result better readable
+          ANSIBLE_CALLBACK_RESULT_FORMAT: yaml
+
+      - name: Commit ansible-build-data and push the changes to github
+        run: |
+         cd "antsibull/build/ansible-build-data/${{ inputs.ansible_major_version }}"
+         git add *.rst *.build *.deps *.yaml *.in validate-tags-ignores
+         git commit -a -m "${{ env.CI_COMMIT_MESSAGE }}"
```
webknjaz, line 93: `-a` doesn't seem to be necessary

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,109 @@
...
+         git push origin "publish-${{ inputs.ansible_version }}"
+
+      - name: Create PR to the ansible-build-data
+        run: |
+          cd antsibull/build/ansible-build-data
+          gh pr create -B main -H "publish-${{ inputs.ansible_version }}" --title "${{ env.CI_COMMIT_MESSAGE }}"  --body "${{ env.CI_COMMIT_MESSAGE }}"
```
webknjaz, line 99: same as in the other PR

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
+name: Upload-ansible-tarball-pypi
+on:
+  workflow_dispatch:
+    inputs:
+      ansible_version:
+        description: 'Release Version. Example : 11.1.0'
+        required: true
+      ansible_major_version:
+        description: 'Exmaple 11'
+        required: true
```
webknjaz, line 10: same typos, kebab-case

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+          python3 -m pip install --upgrade pip
+          python3 -m pip install ansible-core . ../antsibull-core ../antsibull-changelog
+          ansible-galaxy collection install 'git+https://github.com/ansible-collections/community.general.git'
+
+
+      - name: Create sdist dir using the deps file created in the ansible-build-data Release PR
+        run: |
+          antsibull-build rebuild-single "${{ inputs.ansible_version }}" --data-dir "${BUILD_DATA_DIR}" --build-file "${BUILDFILE}" --deps-file "${DEPSFILE}" --sdist-dir antsibull/sdist --debian
+
+      - name: Upload Ansible sdist to PyPI
```
webknjaz, line 80: this is not true. it also uploads wheels

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
+name: Upload-ansible-tarball-pypi
```
webknjaz, line 1: this is not true. it builds both sdist and wheel

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install ansible-core . ../antsibull-core ../antsibull-changelog
+          ansible-galaxy collection install 'git+https://github.com/ansible-collections/community.general.git'
+
+
+      - name: Create sdist dir using the deps file created in the ansible-build-data Release PR
```
webknjaz, line 76: This is not true. The command generates a wheel and an sdist, not just sdist. This is explained @ https://github.com/ansible-community/antsibull/issues/539.

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+  BUILDFILE: "ansible-${{ inputs.ansible_major_version }}.build"
+  DEPSFILE: "ansible-${{ inputs.ansible_version }}.deps"
+
+jobs:
+  publish:
+    name: 'Upload Ansible community distribution (${{ inputs.ansible_version }}) to PyPI'
+    runs-on: ubuntu-latest
+    environment:
+      name: release
+      url: https://test.pypi.org/project/ansible/
```
webknjaz, line 23: make it nicer
```suggestion
      url: https://test.pypi.org/project/ansible/${{ inputs.ansible_version }}
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+  BUILD_DATA_DIR: "antsibull/build/ansible-build-data/${{ inputs.ansible_major_version }}"
+  BUILDFILE: "ansible-${{ inputs.ansible_major_version }}.build"
+  DEPSFILE: "ansible-${{ inputs.ansible_version }}.deps"
+
+jobs:
+  publish:
+    name: 'Upload Ansible community distribution (${{ inputs.ansible_version }}) to PyPI'
+    runs-on: ubuntu-latest
+    environment:
+      name: release
```
webknjaz, line 22: use a more accurate name
```suggestion
      name: testpypi
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,94 @@
...
+      url: https://test.pypi.org/project/ansible/
+    permissions:
+      id-token: write
```
webknjaz, line 25: This privilege mustn't be available to any of the build steps, only to the upload one. It's unacceptable to have it exposed to arbitrarily injected scripts, security-wise. This means that splitting this into separate jobs would be a hard requirement for this PR.

.github/workflows/upload-to-pypi.yml
webknjaz: it's okay to have it if nightly runs are added, I think, to make sure that the pipeline is healthy. the testpypi job would have an unprotected environment, while the real pypi would require a human approval to start the job. but yes, it wouldn't be reasonable to have it in the same job for security reasons (privilege leak) and for structural reasons too (only one environment can be active in a job, meaning that two environments representing two different targets and having different protections set up would need a job each).

anweshadas: First we need to test against test.pypi.org, then in the PR we can move to the production.

anweshadas:
Still working on this, more things left to be updated.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,88 @@
...
+         git add *.rst *.build *.deps *.yaml *.in validate-tags-ignores
+         git commit -m "${{ env.CI_COMMIT_MESSAGE }}"
+         git push origin "publish-${{ inputs.ansible-version }}"
+
+      - name: Create PR to the ansible-build-data
+        working-directory: antsibull/build/ansible-build-data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr create -B main -H "publish-${{ inputs.ansible-version }}" --title "${{ env.CI_COMMIT_MESSAGE }}"  --body "${{ env.CI_COMMIT_MESSAGE }}"
```
webknjaz, line 88: What are these all single-letter CLI args?

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,109 @@
...
+          python3 -m pip install ansible-core antsibull
+          ansible-galaxy collection install community.general
+
+      - name: Create sdist and wheel using the deps file created in the ansible-build-data Release PR
+        working-directory: antsibull
+        run: |
+          ansible-playbook -vv playbooks/build-single-release.yaml -e 'antsibull_ansible-version=${{ inputs.ansible-version }}' -e antsibull_data_reset=false
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
```
webknjaz, line 58: There's a v4 now. It has breaking changes, but since it's the initial integration, it should be fine to just go with it.

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,109 @@
...
+env:
+  BUILD_DATA_DIR: "antsibull/build/ansible-build-data/${{ inputs.ansible-major-version }}"
+  BUILDFILE: "ansible-${{ inputs.ansible-major-version }}.build"
+  DEPSFILE: "ansible-${{ inputs.ansible-version }}.deps"
+
+# Build job builds the wheel and the source tarball and upload the artifacts
+
+jobs:
+  build:
+    name: Build and Upload sdist and wheel
```
webknjaz, line 22: Let's drop "upload", it's confusing to thing that it publishes to PyPI.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,88 @@
...
+        required: true
+      ansible-major-version:
+        description: Example 11
+        required: true
+env:
+  CI_COMMIT_MESSAGE: 'Ansible ${{ inputs.ansible-version }}: Dependencies, changelog and porting guide'
+
+
+jobs:
+  build:
```
webknjaz, line 17: What's the difference between the build job in this workflow and the other one? Why do we need two workflows here?

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,109 @@
...
+    - name: Download artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: sdist-and-wheel
+        path: dist/
+
+    - name: Upload Ansible sdist and wheel to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        repository-url: https://pypi.org/project/ansible
```
webknjaz, line 86: This is an incorrect value, it won't work, because that's not a repository URL but a URL of a project hosted in the index. Delete it, it's not needed at all. The default value contains the correct PyPI upload URL anyway.
```suggestion
```

.github/workflows/ansible-release.yml
anweshadas: `-B` - The branch into which you want your code merged
`-H ` - The branch that contains commits for the pull request

Here is an example PR created by this workflow : `https://github.com/anweshadas/ansible-build-data/pull/10`

.github/workflows/ansible-release.yml
anweshadas: There are two steps right now (in the current process).
- This job creates the .deps, .tags and .yaml files required for the release.
- After the (manual) merging of the PR created by the aforesaid steps we run the second job to build the source tarball and wheel, upload it to PyPI and create the git tag.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    inputs:
+      ansible-version:
+        description: >-
+          Release Version. Example : 11.1.0
+        required: true
+      ansible-major-version:
+        description: Example 11
+        required: true
+env:
+  CI_COMMIT_MESSAGE: 'Ansible ${{ inputs.ansible-version }}: Dependencies, changelog and porting guide'
```
gotmax23, line 13:
```suggestion
  CI_COMMIT_MESSAGE: 'Ansible ${{ inputs.ansible-version }}: Dependencies, changelog and porting guide'
  ANSIBLE_VERSION: '${{ inputs.ansible-version }}'
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install ansible-core antsibull
+          ansible-galaxy collection install community.general
+
+      - name: Checking out to a new branch & setting the user details
+        working-directory: antsibull/build/ansible-build-data
+        run: |
+          git checkout -b "publish-${{ inputs.ansible-version }}"
```
gotmax23, line 54:
```suggestion
          git checkout -b "publish-${ANSIBLE_VERSION}"
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,86 @@
...
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+# Run the playbook according to the current release process
+
+      - name: Test building a release with the defaults
+        working-directory: antsibull
+        run: |
+          ansible-playbook -vv playbooks/build-single-release.yaml \
+          -e antsibull_data_reset=false \
+          -e 'antsibull_ansible_version=${{ inputs.ansible-version }}' \
```
gotmax23, line 65:
```suggestion
          -e "antsibull_ansible_version=${ANSIBLE_VERSION}" \
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,86 @@
...
+          -e 'antsibull_data_dir={{ antsibull_data_git_dir }}/${{ inputs.ansible-major-version }}'
```
gotmax23, line 66: This is already the default.

```suggestion
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,86 @@
...
+
+        env:
+          # Make result better readable
+          ANSIBLE_CALLBACK_RESULT_FORMAT: yaml
+
+      - name: Commit ansible-build-data and push the changes to github
+        working-directory: 'antsibull/build/ansible-build-data/${{ inputs.ansible-major-version }}'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
gotmax23, line 75: You don't need the token to push. It's already set up by the checkout action.

```suggestion
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,86 @@
...
+        run: |
+         git add *.rst *.build *.deps *.yaml *.in validate-tags-ignores
+         git commit -m "${{ env.CI_COMMIT_MESSAGE }}"
+         git push origin "publish-${{ inputs.ansible-version }}"
```
gotmax23, line 79:
```suggestion
         git push origin "publish-${ANSIBLE_VERSION}"
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,86 @@
...
+
+      - name: Create PR to the ansible-build-data
+        working-directory: antsibull/build/ansible-build-data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr create -B main -H "publish-${{ inputs.ansible-version }}" --title "${{ env.CI_COMMIT_MESSAGE }}"  --body "${{ env.CI_COMMIT_MESSAGE }}"
```
gotmax23, line 86: These arguments are redundant anyways, but if you want to keep them, you should use `--head-branch` and `--base-branch` instead of the short options.

```suggestion
          gh pr create --title "${CI_COMMIT_MESSAGE}"  --body "${CI_COMMIT_MESSAGE}"
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,106 @@
...
+          Release Version. Example : 11.1.0
+        required: true
+      ansible-major-version:
+        description: 'Example 11'
+        required: true
+
+env:
+  BUILD_DATA_DIR: "antsibull/build/ansible-build-data/${{ inputs.ansible-major-version }}"
+  BUILDFILE: "ansible-${{ inputs.ansible-major-version }}.build"
+  DEPSFILE: "ansible-${{ inputs.ansible-version }}.deps"
```
gotmax23, line 16:
```suggestion
  DEPSFILE: "ansible-${{ inputs.ansible-version }}.deps"
  ANSIBLE_VERSION: "${{ inputs.ansible-version }}"
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install ansible-core antsibull
+          ansible-galaxy collection install community.general
+
+      - name: Create sdist and wheel using the deps file created in the ansible-build-data Release PR
+        working-directory: antsibull
+        run: |
+          ansible-playbook -vv playbooks/build-single-release.yaml -e 'antsibull_ansible-version=${{ inputs.ansible-version }}' -e antsibull_data_reset=false
```
gotmax23, line 55:
```suggestion
          ansible-playbook -vv playbooks/build-single-release.yaml -e "antsibull_ansible-version=${ANSIBLE_VERSION}" -e antsibull_data_reset=false
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,106 @@
...
+
+# publish job downloads the arifacts and publish it to PyPI
+
+  publish:
+    needs: build
+    name: 'Upload Ansible (${{ inputs.ansible-version }}) to PyPI'
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/project/ansible/${{ inputs.ansible-version }}
```
gotmax23, line 71: Is this the eight environment URL?

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,106 @@
...
+
+    - name: Upload Ansible sdist and wheel to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+
+# git-tag job creates the git tag
+
+  git-tag:
+    needs: publish
+    name: 'Creates git tag for Ansible (${{ inputs.ansible-version }})'
+    runs-on: ubuntu-latest
```
gotmax23, line 91:
```suggestion
    runs-on: ubuntu-latest
    permissions:
      contents: write
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Check out ansible-build-data
+        uses: actions/checkout@v3
+
+      - name: Create git tag
+        working-directory: ansible-build-data
+        run: |
+          git config --global user.name "Github Actions"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git tag -a ${{ inputs.ansible-version }} -m "Ansible ${{ inputs.ansible-version }}: Changelog, Porting Guide and Dependent Collection Details"
```
gotmax23, line 103:
```suggestion
          git tag -a "${ANSIBLE_VERSION}" -m "Ansible ${ANSIBLE_VERSION}: Changelog, Porting Guide and Dependent Collection Details"
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,106 @@
...
+          git push origin ${{ inputs.ansible-version }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
gotmax23, line 106:
```suggestion
```

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,106 @@
+name: Upload-ansible-release-artifacts-to-pypi
+on:
+  workflow_dispatch:
+    inputs:
+      ansible-version:
+        description: >-
+          Release Version. Example : 11.1.0
+        required: true
+      ansible-major-version:
+        description: 'Example 11'
```
gotmax23, line 10:
```suggestion
        description: 'Ansible major version. Example: 11'
```

.github/workflows/upload-to-pypi.yml
gotmax23, line 16: Are these env vars used? It doesn't look like it.

.github/workflows/ansible-release.yml
anweshadas: Replaced it with `gh pr create --base main --head "publish-${{ inputs.ansible-version }}" --title "${{ env.CI_COMMIT_MESSAGE }}"  --body "${{ env.CI_COMMIT_MESSAGE }}"`

.github/workflows/upload-to-pypi.yml
anweshadas: Yes, I was not using them Thank you for pointing it out. Removed it now.

.github/workflows/ansible-release.yml
anweshadas: Edited it to `gh pr create --base main --head "publish-${{ inputs.ansible-version }}" --title "${{ env.CI_COMMIT_MESSAGE }}"  --body "${{ env.CI_COMMIT_MESSAGE }}"`

anweshadas:
Pushed changes based on the feedback.

.github/workflows/ansible-release.yml
webknjaz: Great, thanks! A typical convention for anything resembling bash scripts is using long option names simply because it reduces the mental overhead for the readers, contributing to better maintainability.

.github/workflows/ansible-release.yml
webknjaz: @anweshadas does that sdist need to contain the merged PR contents or can it be built ahead of time?

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install ansible-core antsibull
+          ansible-galaxy collection install community.general
```
webknjaz, line 50: Split these commands into separate steps: having outputs of several processes together is usually hard to follow. And the individual error codes aren't distinguishable either. This recommendation is similar to what we do in Ansible playbooks.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Checking out to a new branch & setting the user details
+        working-directory: antsibull/build/ansible-build-data
+        run: |
+          git checkout -b "publish-${ANSIBLE_VERSION}"
+          git config --global user.name "Github Actions"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
webknjaz, line 57: Setting up Git is also a separate action, let's not mix it with making a new branch. Also, setting user-global config doesn't need changing the directory.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,84 @@
...
+
+# Run the playbook according to the current release process
+
+      - name: Test building a release with the defaults
+        working-directory: antsibull
+        run: |
+          ansible-playbook -vv playbooks/build-single-release.yaml \
+          -e antsibull_data_reset=false \
+          -e "antsibull_ansible_version=${ANSIBLE_VERSION}" \
```
webknjaz, line 66: That last trailing `\` can have a weird side effect that hard to spot. Also, using native YAML syntax, it's possible to produce a single-line command when this YAML file is parsed. With this, you don't need to worry about trailing spaces either, because all the lines will be glued together with whitespaces.
```suggestion
        run: >-
          ansible-playbook -vv playbooks/build-single-release.yaml
          -e antsibull_data_reset=false
          -e "antsibull_ansible_version=${ANSIBLE_VERSION}"
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,84 @@
...
+    steps:
+      - name: Check out antsibull
+        uses: actions/checkout@v3
+        with:
+          repository: ansible-community/antsibull
+          ref: main
+          path: antsibull
+
+      - name: Pre-create build directory
+        run: mkdir -p antsibull/build
```
webknjaz, line 33: Strictly speaking, `mkdir` isn't really necessary, because `git clone` always creates the target directory.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,84 @@
...
+         git add *.rst *.build *.deps *.yaml *.in validate-tags-ignores
+         git commit -m "${{ env.CI_COMMIT_MESSAGE }}"
+         git push origin "publish-${ANSIBLE_VERSION}"
+
+      - name: Create PR to the ansible-build-data
+        working-directory: antsibull/build/ansible-build-data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr create --base main --head "publish-${{ inputs.ansible-version }}" --title "${{ env.CI_COMMIT_MESSAGE }}"  --body "${{ env.CI_COMMIT_MESSAGE }}"
```
webknjaz, line 84: Here, we can use the same "multi-line but really one-line" trick:
```suggestion
        run: >-
          gh pr create
          --base main
          --head "publish-${{ inputs.ansible-version }}"
          --title "${{ env.CI_COMMIT_MESSAGE }}"
          --body "${{ env.CI_COMMIT_MESSAGE }}"
```
isn't it cleaner when flags are paired with their values? I also like using `--<flag>=<value>` style for even better clarity, but that's up to you.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,84 @@
+name: ansible-release
```
webknjaz, line 1: No need to repeat the default value with is the filename. Normally, this field is for human-readable title that shows up in the GH UI.

.github/workflows/upload-to-pypi.yml
webknjaz: @gotmax23 what do you mean? The link looks correct to me.

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Check out ansible-build-data
+        uses: actions/checkout@v3
+
+      - name: Create git tag
+        working-directory: ansible-build-data
+        run: |
+          git config --global user.name "Github Actions"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git tag -a "${ANSIBLE_VERSION}" -m "Ansible ${ANSIBLE_VERSION}: Changelog, Porting Guide and Dependent Collection Details"
+          git push origin ${{ inputs.ansible-version }}
```
webknjaz, line 104: Shouldn't this also use `"${ANSIBLE_VERSION}"`?

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Check out ansible-build-data
+        uses: actions/checkout@v3
+
+      - name: Create git tag
+        working-directory: ansible-build-data
+        run: |
+          git config --global user.name "Github Actions"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git tag -a "${ANSIBLE_VERSION}" -m "Ansible ${ANSIBLE_VERSION}: Changelog, Porting Guide and Dependent Collection Details"
```
webknjaz, line 103: Looks like this tags whatever's the last commit on the main branch. So if there's a race condition with something that manages squeeze in after the PR merge but before running this workflow, some other commit may end up being tagged, not the immediate PR merge. To combat this, we should probably track the merge commit associated with the PR.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,84 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      ansible-version:
+        description: >-
+          Release Version. Example : 11.1.0
+        required: true
+      ansible-major-version:
+        description: Example 11
+        required: true
```
webknjaz, line 11: This input isn't really needed. Instead, calculate the major version from `ansible-version` and set that as a step/job output for reuse in runtime.

webknjaz:
@anweshadas I see that you made 2 workflows with separate triggers and most of the process being the same. Do I understand correctly, that it's one process split in half with a pause/gate in the middle?

It looks like this doesn't really need a separate trigger. Instead, I'd make one main workflow with `workflow_dispatch` and put parts into reusable workflows with `workflow_call`. The reusable workflows would be "called" (or "included") from that main one. And the workflow run would include all the bits. The job happening after the PR creation would be marked with some environment (like `await-pr-merge`) and with required reviews for that environment (configured in the repository settings), it'd pause the process. Once the PR's merged, you'd click a button in the paused workflow run to make it proceed.

With that, you could even generate dists together with making the PR, save them as workflow artifacts and only perform upload+tagging after unpausing. And since it's a single trigger, you'd have to type in the version just once.

.github/workflows/ansible-release.yml
anweshadas: > @anweshadas does that sdist need to contain the merged PR contents or can it be built ahead of time?

The `sdist` should not be build prior to the merging of the PR as we verify the `deps` file during the PR review.

.github/workflows/upload-to-pypi.yml
anweshadas: Yes, this will be a future improvement.

anweshadas:
> @anweshadas I see that you made 2 workflows with separate triggers and most of the process being the same. Do I understand correctly, that it's one process split in half with a pause/gate in the middle?


I will learn about this and about this in the upcoming PRs.

.github/workflows/ansible-release.yml
webknjaz: Why?

webknjaz: I mean, what's missing? Can't it be built anyway and discarded if the review result wasn't acceptable?

.github/workflows/ansible-release.yml
anweshadas: Edited the as per the suggestion.

.github/workflows/ansible-release.yml
anweshadas: Edited the as per the suggestion.

.github/workflows/ansible-release.yml
anweshadas: Edited the as per the suggestion.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,90 @@
...
+          python3 -m pip install antsibull
+          ansible-galaxy collection install community.general
+
+      - name: Checking out to a new branch
+        working-directory: antsibull/build/ansible-build-data
+        run: |
+          git checkout -b "publish-${ANSIBLE_VERSION}"
+
+      - name: Setting the user details
+        working-directory: antsibull/build/ansible-build-data
```
webknjaz, line 59: No need to chdir, as mentioned in https://github.com/ansible-community/ansible-build-data/pull/265/files#r1442368714
```suggestion
      - name: Configure user-global Git author and committer metadata
```

.github/workflows/ansible-release.yml
webknjaz: Looks like you forgot to commit this?

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install ansible-core antsibull
+          ansible-galaxy collection install community.general
+
+      - name: Create sdist and wheel using the deps file created in the ansible-build-data Release PR
+        working-directory: antsibull
+        run: |
+          ansible-playbook -vv playbooks/build-single-release.yaml -e "antsibull_ansible-version=${ANSIBLE_VERSION}" -e antsibull_data_reset=false
```
webknjaz, line 53: typo?
```suggestion
          ansible-playbook -vv playbooks/build-single-release.yaml -e "antsibull_ansible_version=${ANSIBLE_VERSION}" -e antsibull_data_reset=false
```

gotmax23:
I agree with @webknjaz here. If possible, I'd really like to see the dists uploaded as GHA artifacts that can be reviewed and tested as part of the PR and then have those same artifacts uploaded to PyPI after the PR is merged instead of building them twice.

.github/workflows/upload-to-pypi.yml
anweshadas: Yes, updated.

.github/workflows/ansible-release.yml
anweshadas: Updated.

anweshadas:
@webknjaz @gotmax23 I updated the PR based on the review. These are the following changes:

1. Added a new job to pause the workflow, created a new environment: `pypirelease` for that .
2. Added `release-managers` as reviewers, could not add `ansible-community/release-management-wg`. (Will look into it why)
3. Instead of 2 workflows (a. creating the sdist & wheel and b. uploading it to PyPI & creating the git tag) we will have one workflow. If the changes look good  I will remove the other file (`upload-to-pypi.yml`).

gotmax23:
> Added release-managers as reviewers, could not add ansible-community/release-management-wg. (Will look into it why)

You probably have to add that group as maintainers of this repo first.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,152 @@
...
+      - name: Pause for the PR review
+        run: echo 'Review Done'
+
+  publish:
+    needs: [build, prreview]
+    name: 'Upload Ansible (${{ inputs.ansible-version }}) to PyPI'
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/project/ansible
```
gotmax23, line 116: Why not

```suggestion
    needs: [build]
    name: 'Upload Ansible (${{ inputs.ansible-version }}) to PyPI'
    runs-on: ubuntu-latest
    environment:
      name: pypirelease
      url: https://pypi.org/project/ansible
```

and then remove the extra job?

cybette:
Added `release-management-wg` team as maintainers

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+          --body "${{ env.CI_COMMIT_MESSAGE }}"
+
+  # publish job downloads the arifacts and publish it to PyPI
+
+  publish:
+    needs: [build]
+    name: 'Upload Ansible (${{ inputs.ansible-version }}) to PyPI'
+    runs-on: ubuntu-latest
+    environment:
+      name: pypirelease
```
webknjaz, line 107: Why not just
```suggestion
      name: pypi
```
?

A GitHub environment corresponds to the deployment target (a place — PyPI), not the process (release).

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+          gh pr create
+          --base main
+          --head "publish-${{ inputs.ansible-version }}"
+          --title "${{ env.CI_COMMIT_MESSAGE }}"
+          --body "${{ env.CI_COMMIT_MESSAGE }}"
+
+  # publish job downloads the arifacts and publish it to PyPI
+
+  publish:
+    needs: [build]
```
webknjaz, line 103: Since it's just one job, let's not make it a list:
```suggestion
    needs: build
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+      - name: Create PR to the ansible-build-data
+        working-directory: antsibull/build/ansible-build-data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: >-
+          gh pr create
+          --base main
+          --head "publish-${{ inputs.ansible-version }}"
+          --title "${{ env.CI_COMMIT_MESSAGE }}"
+          --body "${{ env.CI_COMMIT_MESSAGE }}"
```
webknjaz, line 98: I think it's confusing to use the same text as both title and description of the PR. Can we separate them by context/semantics?

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+
+  # publish job downloads the arifacts and publish it to PyPI
+
+  publish:
+    needs: [build]
+    name: 'Upload Ansible (${{ inputs.ansible-version }}) to PyPI'
+    runs-on: ubuntu-latest
+    environment:
+      name: pypirelease
+      url: https://pypi.org/project/ansible
```
webknjaz, line 108: Since we know the version at this point, let's make the URL more accurate:
```suggestion
      url: https://pypi.org/project/ansible/${{ inputs.ansible-version }}
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+
+      - name: Create PR to the ansible-build-data
+        working-directory: antsibull/build/ansible-build-data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: >-
+          gh pr create
+          --base main
+          --head "publish-${{ inputs.ansible-version }}"
+          --title "${{ env.CI_COMMIT_MESSAGE }}"
```
webknjaz, line 97: Perhaps
```suggestion
          --title "Release Ansible v${{ inputs.ansible-version }}"
```
would do.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+      ansible-version:
+        description: >-
+          Release Version. Example : 11.1.0
+        required: true
+      ansible-major-version:
+        description: Example 11
+        required: true
+env:
+  CI_COMMIT_MESSAGE: 'Ansible ${{ inputs.ansible-version }}: Dependencies, changelog and porting guide'
+  ANSIBLE_VERSION: '${{ inputs.ansible-version }}'
```
webknjaz, line 14:
```suggestion
  ANSIBLE_VERSION: ${{ inputs.ansible-version }}
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+
+jobs:
+  build:
+    name: 'Build Ansible (${{ inputs.ansible-version }})'
```
webknjaz, line 18:
```suggestion
    name: Build Ansible (${{ inputs.ansible-version }})
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+          --base main
+          --head "publish-${{ inputs.ansible-version }}"
+          --title "${{ env.CI_COMMIT_MESSAGE }}"
+          --body "${{ env.CI_COMMIT_MESSAGE }}"
+
+  # publish job downloads the arifacts and publish it to PyPI
+
+  publish:
+    needs: [build]
+    name: 'Upload Ansible (${{ inputs.ansible-version }}) to PyPI'
```
webknjaz, line 104:
```suggestion
    name: Upload Ansible (${{ inputs.ansible-version }}) to PyPI
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+        path: dist/
+
+    - name: Upload Ansible sdist and wheel to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+
+  # git-tag job creates the git tag
+
+  git-tag:
+    needs: publish
+    name: 'Creates git tag for Ansible (${{ inputs.ansible-version }})'
```
webknjaz, line 127:
```suggestion
    name: Creates git tag for Ansible (${{ inputs.ansible-version }})
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+    inputs:
+      ansible-version:
+        description: >-
+          Release Version. Example : 11.1.0
+        required: true
+      ansible-major-version:
+        description: Example 11
+        required: true
+env:
+  CI_COMMIT_MESSAGE: 'Ansible ${{ inputs.ansible-version }}: Dependencies, changelog and porting guide'
```
webknjaz, line 13:
```suggestion
  CI_COMMIT_MESSAGE: >-
    Ansible ${{ inputs.ansible-version }}:
    Dependencies, changelog and porting guide
```

webknjaz (commented):
(nitpick) Let's the unnecessary YAML quotes that are added added inconsistently in arbitrary places.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+      # This is where the antsibull build-release role expects it by default
+      - name: Check out ansible-build-data under antsibull build directory
+        uses: actions/checkout@v3
+        with:
+          path: antsibull/build/ansible-build-data
+
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
```
webknjaz, line 44: (one more)
```suggestion
          python-version: 3.11
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+    needs: publish
+    name: 'Creates git tag for Ansible (${{ inputs.ansible-version }})'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+
+      - name: Check out ansible-build-data
+        uses: actions/checkout@v3
```
webknjaz, line 135: I think, this needs an explicit `ref` input set. By default, it'll probably target the commit attached to the original `workflow_dispatch` event that triggered this workflow run. However, we need to tag the merge commit after the PR, not the commit before the PR, since we want the tagged version to contain changes from the PR, right?

webknjaz: If you want to postpone this — it's fine. But then, you need to at least force this specific checkout step to fetch the top of the main branch, not some old pre-PR commit:
```suggestion
        uses: actions/checkout@v3
        with:
          ref: ${{ github.event.repository.default_branch }}
```
(and implement what I suggested above in a follow-up PR)

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+
+      - name: Create git tag
```
webknjaz, line 137: May I suggest also creating a GitHub Release? These tend to show up nicer on GH and are still associated with tags.

webknjaz:
@anweshadas another review round complete. Part of the comments are style normalization — you can safely accept the suggested changes in a batch from the Files tab in this PR.
And there's a few comments pointing out more serious problems that I believe should be addressed before merging this PR.

.github/workflows/ansible-release.yml
anweshadas: > To implement this, we need to save the PR number as an output of the `build` job above. Then, we'd access that PR number in this job, and query GitHub API (or maybe just use GH CLI, not sure) to retrieve the PR metadata and see which commit exactly is the merge commit associated with that PR.


I will file a separate PR on this context as a future enhancement.

.github/workflows/ansible-release.yml
anweshadas: This is also a new feature (enhancement), I will file a separate on this in future.

.github/workflows/ansible-release.yml
anweshadas: Updated.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,150 @@
...
+      - name: Create PR to the ansible-build-data
+        working-directory: antsibull/build/ansible-build-data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: >-
+          gh pr create
+          --base main
+          --head "publish-${{ inputs.ansible-version }}"
+          --title "Release Ansible v${{ inputs.ansible-version }}"
+          --body "${{ env.CI_COMMIT_MESSAGE }}"
```
gotmax23, line 101:
```suggestion
          --head "publish-${ANSIBLE_VERSION}"
          --title "Release Ansible v${ANSIBLE_VERSION}"
          --body "${CI_COMMIT_MESSAGE}"
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,150 @@
...
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+
+      - name: Create git tag
+        working-directory: ansible-build-data
+        run: |
+          git config --global user.name "Github Actions"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git tag -a "${ANSIBLE_VERSION}" -m "Ansible ${ANSIBLE_VERSION}: Changelog, Porting Guide and Dependent Collection Details"
+          git push origin ${ANSIBLE_VERSION}}
```
gotmax23, line 149:
```suggestion
          git push origin "${ANSIBLE_VERSION}"
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,150 @@
...
+
```
gotmax23, line 150:
```suggestion
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,150 @@
...
+        working-directory: antsibull
+        run: >-
+          ansible-playbook -vv playbooks/build-single-release.yaml
+          -e antsibull_data_reset=false
+          -e "antsibull_ansible_version=${ANSIBLE_VERSION}"
+
+        env:
+          # Make result better readable
+          ANSIBLE_CALLBACK_RESULT_FORMAT: yaml
+
```
gotmax23, line 77:
```suggestion
```

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,144 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.11'
+
+      - name: Install dependencies
+        working-directory: antsibull
+        run: |
+          python3 -m pip install ansible-core
+          python3 -m pip install antsibull
+          ansible-galaxy collection install community.general
```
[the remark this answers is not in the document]
gotmax23: Actually, I think we already have one.

gotmax23, line 53:
```suggestion
          ansible-galaxy install -r requirements.yml
```

.github/workflows/ansible-release.yml
felixfontein: The current version of this job is uploading both sdist and wheel.

felixfontein:
We tried to apply most of the feedback. PTAL.

.github/workflows/ansible-release.yml
webknjaz: @felixfontein I think my initial concern was regarding building the artifacts (both), not uploading (which is to be a separate thing).

.github/workflows/ansible-release.yml
webknjaz: Okay, fair. Though, it's just another interface to making a tag so it might be viewed as invoking a different command rather than adding a new feature. But that's up to you, of course.

.github/workflows/ansible-release.yml
webknjaz: @anweshadas thanks! Did you also update it in the repo settings?

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,150 @@
...
+        run: |
+          git config --global user.name "Github Actions"
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      # Run the playbook according to the current release process
+
+      - name: Building a release with the defaults
+        working-directory: antsibull
+        run: >-
+          ansible-playbook -vv playbooks/build-single-release.yaml
```
webknjaz, line 71: @gotmax23 so I have a comment for you which I believe might need to be implemented in `antsibull`. This playbook builds the dists but doesn't perform the metadata validation. That must happen early in the process. Specifically, calling `twine check --strict` must succeed. My `pypi-upload` action is invoked late in the process and it does perform a non-strict check, which may fail the upload last moment which would be unpleasant. The PyPI backend might also reject uploads because of the problems with dists, some of which (but not all) can be caught in the strict mode.
Running the check as early as possible (on merges to `main` and pull requests) reveals metadata problems in a way that they can be fixed before attempting to build though this workflow.

So this workflow is the last line of defense that may reveal problems before the PR is created. I don't know if there's currently other places where it would be possible to inject this check, but perhaps putting it into that playbook would be useful. Something that could abort a release with broken meta.

gotmax23: Yeah, adding it to the playbook (or _maybe_ shelling out to it from the python code) definitely makes sense to me. Can you open an issue in ansible's issue tracker?

webknjaz: Done: #577. Marking this thread as resolved.

webknjaz (approved):
You have my blessing, I suppose..

gotmax23:
Requesting a review from myself. I'll give this another look.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,150 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: sdist-and-wheel
+          path: antsibull/build/ansible-*.*
+
+      - name: Commit ansible-build-data and push the changes to github
+        working-directory: >-
+          antsibull/build/ansible-build-data/${{ inputs.ansible-major-version }}
+        run: |
+         git add *.rst *.build *.deps *.yaml *.in validate-tags-ignores
```
gotmax23, line 88: I would prefer a `git add .` here. I don't want to have to have to adjust this in the future if we add additional files.

.github/workflows/ansible-release.yml
```diff
@@ -0,0 +1,150 @@
...
+         git commit -m "${{ env.CI_COMMIT_MESSAGE }}"
```
gotmax23, line 89:
```suggestion
         git commit -m "${CI_COMMIT_MESSAGE}"
```

gotmax23:
Other than that, this looks good to me. Thanks @anweshadas for all your work on this and for responding to all the feedback! This is really great.

.github/workflows/ansible-release.yml
felixfontein:
```suggestion
         git add .
```

felixfontein (approved):
Let's get this merged so we can do further improvements in follow-up PRs :)

felixfontein:
CI failures are unrelated.

felixfontein:
Thanks @anweshadas @gotmax23 @webknjaz for your work on this!

PR #271, opened 2023-07-18
Deprecate netapp.azure

mariolenz:
Deprecate netapp.azure as discussed in ansible-community/community-topics#234 and voted on in ansible-community/community-topics#249


mariolenz:
Thanks @gotmax23!

PR #274, opened 2023-07-25
9: add inspur.ispim to validate-tags-ignore

gotmax23:
We don't currently have a way to pin collection versions for prior to
X.Y.0a1, so we'll ignore it for now.

Relates: https://github.com/ispim/inspur.ispim/issues/24


baoshanwang:
I've added a 2.0.1 tag to inspur.ispim.

gotmax23:
Cool, thanks!

PR #275, opened 2023-08-01
Deprecate netapp.elementsw

mariolenz:
Deprecate netapp.elementsw as discussed in ansible-community/community-topics#235 and voted on in ansible-community/community-topics#250


mariolenz:
Thanks for reviewing @gotmax23!

PR #278, opened 2023-08-04
Adds execution-environment.yml required for ee build and release

anweshadas:


gotmax23:
I'm a bit confused. Was this announced? Why is this part of the ansible-build-data repository when it's built separately and missing most of the community collections? https://github.com/ansible-community/images would be a better place for this as @felixfontein said in the other ticket.

Andersson007:
@gotmax23 nope, it wasn't announced but it will be soon.

>Why is this part of the ansible-build-data repository

The plan is to release them together. Please wait for the announcement.

>and missing most of the community collections?

It doesn't make sense to package most of the community collections. See the [EE getting started](https://docs.ansible.com/ansible/devel/getting_started_ee/index.html) for details how EE work.
There's intentionally very basic set of them. If they have `ansible.*` in names, doesn't mean there cannot be community maintainers / contributors. They live under the ansible_collections org. And it's even not really important which ones we have there as the power of EE unfolds when you build images yourself.
We can redefine things later.
The goal of the image is to provide a very basic image for users to play with (including for the purpose of the EE getting started guide). If we put all the collections, there will be conflicts.

That image is definitely not supposed to become a killer feature of Ansible but rather a tool for entry threshold mitigation or for users who don't need much of content besides `builtin` or just want to run something quickly out of curiosity.

8/execution-environment.yml
```diff
@@ -0,0 +1,23 @@
+----
+
+version: 3
+images:
+  base_image:
+    name: quay.io/fedora/fedora:latest
+dependencies:
+  ansible_core:
+    package_pip: ansible-core==2.15.1
```
Ompragash, line 9: EE's are Core specific and not "Ansible Community Package" specific. I believe it would be more suitable to include them in the [ansible-community/images](https://github.com/ansible-community/images) repository instead of here.

I understand the reasoning behind having them here to facilitate the building and releasing of both the Community Package and EE together. However, consolidating all image files under one roof (i.e., in [ansible-community/images](https://github.com/ansible-community/images)) would be highly beneficial from a community perspective.

gotmax23:
> The plan is to release them together.

IMO, releasing them together will only lead to confusion. The ansible package contains a specific set of collections with specific versions. This EE is something different.

> The goal of the image is to provide a very basic image for users to play with (including for the purpose of the EE getting started guide).

That sounds like a great idea. I am just opposed to tying it to the community package.

mariolenz:
> > The plan is to release them together.
> 
> IMO, releasing them together will only lead to confusion. The ansible package contains a specific set of collections with specific versions. This EE is something different.

I agree, this EE doesn't look related to the ansible (community) package. So why have it here in this repository that, as far as I understand, is about releasing the community package?

> > The goal of the image is to provide a very basic image for users to play with (including for the purpose of the EE getting started guide).
> 
> That sounds like a great idea. I am just opposed to tying it to the community package.

I'm with you. And I agree with @Ompragash: Why not have it directly in [ansible-community/images](https://github.com/ansible-community/images)?

Andersson007:
thanks for the great feedback, folks!
i even didn't know that https://github.com/ansible-community/images exists.. thanks, SGTM
@felixfontein i see you committed recently, would you like to 1. suggest a location for the file 2. or raise the PR?

felixfontein:
@Andersson007 it should be added into a subdirectory of https://github.com/ansible-community/images/tree/main/execution-environments, and added to the EE tests (https://github.com/ansible-community/images/blob/main/.github/workflows/execution-environments.yml). Right now there is no Zuul code to rebuild and publish the EEs, there's only something for the ansible-test images. I don't know whether the credentials included in the Zuul config have access rights to publish to https://quay.io/ansible-community/test-ee:xxx, which is where the current EE (or let's better say "EE", as it's not really a proper EE) is published to.

Andersson007:
@felixfontein thanks for clarifying! cc @anweshadas ^

anweshadas:
> @felixfontein thanks for clarifying! cc @anweshadas ^

After I open the new PR I will close this one. Thank you everyone for the feedback.

8/execution-environment.yml
samccann: by the time this comes out, I think it will be core 2.15.2 that we'd want? (to match the most recent core release?)

Is there a plan to update this ee for every cor release or just...periodically (aka randomly ;-)

anweshadas:
https://github.com/ansible-community/images/pull/61 created. Closing this PR.

PR #284, opened 2023-08-29
Deprecate hpe.nimble

mariolenz:
Deprecate hpe.nimble as discussed in ansible-community/community-topics#254 and voted on in ansible-community/community-topics#265.


felixfontein:
For the record, I approved this via IRC earlier today.

Issue #324, opened 2023-11-12
Document the release process for X.Y.0 freeze releases


gotmax23:
See the instructions in https://github.com/ansible-community/ansible-build-data/issues/211#issuecomment-1567650513 and https://github.com/ansible-community/ansible-build-data/pull/179#issuecomment-1324314793. This should be completed prior to the 2023-11-21 9.0.0 release.

From https://github.com/ansible-community/ansible-build-data/issues/221:

> `docs/new-ansible.md` - explanation about creating a new ansible release. This should cover creating the `{MAJOR VERSION}` directory, generating the `ansible-{MAJOR VERSION}.build` file, and creating frozen releases. This would serve as an appendix to `docs/release-process.md`

Issue #221, opened 2023-04-26
Improve release process documentation


felixfontein:
Create a folder `docs/`, and create documents in there describing all aspects of the release process.

gotmax23:
https://github.com/ansible-community/ansible-build-data/pull/224 is a start. Feedback welcome!

Andersson007:
SGTM

gotmax23:
I took a stab at documenting the release process (`docs/release-process.md`) in https://github.com/ansible-community/ansible-build-data/pull/227.

mariolenz:
Can we close this? I mean, I've done some releases following the documentation that's been provided up until now. The automated release workflow is really helpful, but even the documentation on manually releasing looks god to me.

So can we close this, or are there any reasons to keep this open?

felixfontein:
I think this is done. As always documentation can always be improved, but the current state is good enough for now I think :)

PR #325, opened 2023-11-12
docs: add new-ansible-and-freezes documentation

gotmax23:
Fixes: https://github.com/ansible-community/ansible-build-data/issues/324
Relates: https://github.com/ansible-community/ansible-build-data/issues/221

### DIFF
--- a/docs/new-ansible-and-freezes.md
+++ b/docs/new-ansible-and-freezes.md
@@ -14,7 +14,7 @@ see [Ansible Release Process](release-process.md).
 After the release of `X.0.0`, it is necessary to create the directory
 structure for Ansible `X+1`.
 
-1. Determine the current major version and previous major version
+1. Determine the current major version and next major version
 
     ``` sh
     CURRENT_MAJOR_VERSION=9
--- a/docs/new-ansible-and-freezes.md
+++ b/docs/new-ansible-and-freezes.md
@@ -1,0 +1,149 @@
+# New Ansible Releases and Freezes
+
+## Preamble
+
+Releasing new Ansible major versions and frozen releases requires some special
+handling.
+For information about the general release process,
+see [Ansible Release Process](release-process.md).
+
+## Setting up for a new major release
+
+<!-- TODO: Write a script, playbook, or antsibull-build subcommand to automate this -->
+
+After the release of `X.0.0`, it is necessary to create the directory
+structure for Ansible `X+1`.
+
+1. Determine the current major version and previous major version
+
+    ``` sh
+    CURRENT_MAJOR_VERSION=9
+    NEXT_MAJOR_VERSION=10
+    ```
+
+2. Create the major version directory
+
+    ``` sh
+    mkdir "${NEXT_MAJOR_VERSION}/"
+    ```
+
+3. Copy over the `ansible.in` and `collection-meta.yaml` files
+
+    ``` sh
+    cp "${CURRENT_MAJOR_VERSION}/ansible.in" "${CURRENT_MAJOR_VERSION}/collection-meta.yaml" \
+       "${NEXT_MAJOR_VERSION}/"
+    ```
+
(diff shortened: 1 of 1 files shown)


docs/new-ansible-and-freezes.md
```diff
@@ -0,0 +1,146 @@
...
+see [Ansible Release Process](release-process.md).
+
+## Setting up for a new major release
+
+<!-- TODO: Write a script, playbook, or antsibull-build subcommand to automate this -->
+
+After the release of `X-1.0.0`, it is necessary to create the directory
+structure for Ansible `X`.
+
+1. Determine the current major version and previous major version
```
felixfontein, line 17: I think current would mean X-1. Right now the current major version is 8, and will change to 9 on Tuesday when 9.0.0 is released. But we want to create the 10 directory, so we should say 'next major version' IMO.
```suggestion
1. Determine the next major version and current major version
```

docs/new-ansible-and-freezes.md
```diff
@@ -0,0 +1,146 @@
...
+
+    ``` sh
+    MAJOR_VERSION=10
+    PREVIOUS_MAJOR_VERSION=9
```
felixfontein, line 21: Maybe change these to `NEXT_MAJOR_VERSION` and `CURRENT_MAJOR_VERSION`?

docs/new-ansible-and-freezes.md
```diff
@@ -0,0 +1,146 @@
...
+
+    ``` sh
+    cp "${PREVIOUS_MAJOR_VERSION}/ansible.in" "${PREVIOUS_MAJOR_VERSION}/collection-meta.yaml" \
+       "${MAJOR_VERSION}/"
+    ```
+
+4. Symlink `X-1.0.0`'s deps file to `${MAJOR_VERSION}/ancestor.deps`
+
+    ``` sh
+    ln -sr "${PREVIOUS_MAJOR_VERSION}/${PREVIOUS_MAJOR_VERSION}.0.0.deps" \
```
felixfontein, line 40:
```suggestion
    ln -sr "${PREVIOUS_MAJOR_VERSION}/ansible-${PREVIOUS_MAJOR_VERSION}.0.0.deps" \
```

docs/new-ansible-and-freezes.md
```diff
@@ -0,0 +1,146 @@
...
+    antsibull-build validate-tags \
+        --data-dir . \
+        --ignores-file validate-tags-ignores \
+        --output "ansible-${VERSION}-tags.yaml" \
+        "${VERSION}"
+    ```
+
+8. Run the the release playbook as outlined in [Ansible Release Process](release-process.md).
+   Make sure to use pass `-e antsibull_data_reset=false` to preserve the
+   ansible-build-data modifcations.
```
felixfontein, line 146: The release playbook runs `antsibull-build prepare` anyway, which adds a changelog fragment (so step 5 isn't needed) and overwrites the `.deps` and `.yaml` files created in step 4 and 6, and also the tags data file from step 7.

gotmax23: The playbook is idempotent, and it will not run those steps if the `.deps` file is present. See https://github.com/ansible-community/antsibull/blob/9cb292752349bdc5f1f5db0cadcf0f61e62e1de2/roles/build-release/tasks/build.yaml#L60.

docs/new-ansible-and-freezes.md
gotmax23: Good point. I'll change those.

gotmax23:
@felixfontein, thanks for the review! PTAL.

docs/new-ansible-and-freezes.md
felixfontein: Oh right, I missed that... Sorry!

felixfontein (approved):

gotmax23:
Thanks!

PR #341, opened 2023-12-05
mkdocs.yml: fix copyright footer

gotmax23:
The license is GPLv3 *or later,* and we should link to our own LICENSE
file, not the one in core.


felixfontein:
@gotmax23 thanks for fixing this!

PR #353, opened 2024-01-21
Ansible 7 and 8 are not compatible with community.docker 3.6.0 due to a new dependency

felixfontein:
While these versions are EOL, at least Ansible 7 is still used in antsibull-core's CI.


gotmax23 (approved):
Thanks! I was going to ping you about this but looks like you beat me to it :).

PR #367, opened 2024-02-11
docs release-process: use GH_USERNAME as a placeholder

gotmax23:
At least on my system, the UNIX username is different than my Github
username. Let's change this var name accordingly so it can be more
easily copy-pasted.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -53,8 +53,8 @@ This only needs to be done once.
     This guide uses your Github username as the fork remote name.
 
     ```
-    git remote add ${USERNAME} https://github.com/${USERNAME}/ansible-build-data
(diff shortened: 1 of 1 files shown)


felixfontein (approved):

webknjaz:
Fair, I have a different OS login too, though I also tend to fork repos adding the upstream org as a prefix. That's another corner case that's not evident. Does this doc really need to include every quirk? All such things are pretty much person-dependent. I'd say that the doc should describe the process that works for the majority, and the power users can figure it out.
Another improvement is automating these commands. It's not always possible to guess the right remote, but there could be a config with a mapping, like cherry-picker has.

ssbarnea (approved):

PR #387, opened 2024-04-17
Prune community.sap from Ansible 10

mariolenz:
This collection is starting to haunt me. We've removed it (for the second time!) in #360. But it still turns up in some files :tired_face:


felixfontein (changes requested):
community.sap was renamed to community.sap_libs. You're now trying to remove the new name as well :)

mariolenz:
Damn! You're right. I've grep'ed for `comunity.sap` and didn't see that the matches I've found were `community.sap_libs` :facepalm:

PR #390, opened 2024-04-23
Improve release workflow

felixfontein:
The first commit stores the PR's URL and adds a step before uploading the release to PyPI which checks whether the PR has been merged. If it hasn't been merged, the `publish` stage fails. (This isn't the best solution possible, but will prevent something like the accidental 9.5.0 release from today.)

The second commit adds a Python step that parses the package version to validate it, and to extract the major version. This allows to simplify the workflow's input to a single version field.

### DIFF
--- a/.github/workflows/ansible-release.yml
+++ b/.github/workflows/ansible-release.yml
@@ -6,10 +6,6 @@ on:
         description: >-
           Release Version. Example: 11.1.0
         required: true
-      ansible-major-version:
-        description: >-
-          Major Release Version. Example: 11
-        required: true
 env:
   CI_COMMIT_MESSAGE: >-
     Ansible ${{ inputs.ansible-version }}:
@@ -51,9 +47,38 @@ jobs:
       - name: Install dependencies
         working-directory: antsibull
         run: |
+          python3 -m pip install packaging
           python3 -m pip install ansible-core
           python3 -m pip install antsibull
           ansible-galaxy install -r requirements.yml
+
+      - name: Validate version and extract major version
+        shell: python
+        id: extract-version
+        run: |
+          import os
+          import pathlib
+          import sys
+
+          from packaging.version import Version
+
+          FILE_APPEND_MODE = 'a'
+          OUTPUTS_FILE_PATH = pathlib.Path(os.environ['GITHUB_OUTPUT'])
+          VERSION = os.environ['ANSIBLE_VERSION']
+
+          def set_output(name, value):
+              with OUTPUTS_FILE_PATH.open(FILE_APPEND_MODE) as outputs_file:
+                  outputs_file.writelines(f'{name}={value}{os.linesep}')
+
+          try:
+              version = Version(VERSION)
+          except Exception as exc:
+              print(
+                  f'::error ::The version {VERSION!r} cannot be parsed: {exc}.'
+              )
+              sys.exit(1)
+
+          set_output('major-version', version.major)
 
       - name: Checking out to a new branch
         working-directory: antsibull/build/ansible-build-data
@@ -86,7 +111,7 @@ jobs:
 
       - name: Commit ansible-build-data and push the changes to github
         working-directory: >-
-          antsibull/build/ansible-build-data/${{ inputs.ansible-major-version }}
+          antsibull/build/ansible-build-data/${{ steps.extract-version.outputs.major-version }}
         run: |
          git add .
          git commit -m "${CI_COMMIT_MESSAGE}"
--- a/.github/workflows/ansible-release.yml
+++ b/.github/workflows/ansible-release.yml
@@ -23,6 +23,8 @@ jobs:
     permissions:
       pull-requests: write
       contents: write
+    outputs:
+      pr_url: ${{ steps.create-pr.outputs.pr_url }}
 
     steps:
       - name: Check out antsibull
@@ -91,17 +93,19 @@ jobs:
          git push origin "publish-${ANSIBLE_VERSION}"
 
       - name: Create PR to the ansible-build-data
+        id: create-pr
         working-directory: antsibull/build/ansible-build-data
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           ARTIFACT_URL: ${{ steps.upload-artifact.outputs.artifact-url }}
         run: |
           body="$(echo -e "${CI_COMMIT_MESSAGE}\nRelease artifacts: <${ARTIFACT_URL}>")"
+          echo -n "pr_url=" >> "$GITHUB_OUTPUT"
           gh pr create \
           --base main \
           --head "publish-${ANSIBLE_VERSION}" \
           --title "Release Ansible ${ANSIBLE_VERSION}" \
-          --body "${body}"
+          --body "${body}" >> "$GITHUB_OUTPUT"
 
   # publish job downloads the arifacts and publish it to PyPI
 
@@ -116,6 +120,16 @@ jobs:
       id-token: write
 
     steps:
+
+    - name: Ensure that the PR was merged
+      env:
+        PR_URL: ${{ needs.build.outputs.create-pr }}
+      run: |
+        STATE=$(gh pr view "${PR_URL}" --json state --template "{{.state}}")
+        if [ "${STATE}" != "MERGED" ]; then
+          echo "::error ::The state of PR ${PR_URL} must be MERGED, not ${STATE}"
+          exit 1
+        fi
 
     - name: Download artifact
       uses: actions/download-artifact@v4
--- a/.github/workflows/ansible-release.yml
+++ b/.github/workflows/ansible-release.yml
@@ -73,10 +73,9 @@ jobs:
           try:
               version = Version(VERSION)
           except Exception as exc:
-              print(
+              sys.exit(
                   f'::error ::The version {VERSION!r} cannot be parsed: {exc}.'
               )
-              sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ansible-release.yml
```diff
@@ -49,10 +47,39 @@ jobs:
...
+        shell: python
+        id: extract-version
+        run: |
+          import os
+          import pathlib
+          import sys
+
+          from packaging.version import Version
+
+          FILE_APPEND_MODE = 'a'
```
gotmax23, line 65: Does this really need to be a constant?

.github/workflows/ansible-release.yml
```diff
@@ -49,10 +47,39 @@ jobs:
...
+          OUTPUTS_FILE_PATH = pathlib.Path(os.environ['GITHUB_OUTPUT'])
+          VERSION = os.environ['ANSIBLE_VERSION']
+
+          def set_output(name, value):
+              with OUTPUTS_FILE_PATH.open(FILE_APPEND_MODE) as outputs_file:
+                  outputs_file.writelines(f'{name}={value}{os.linesep}')
```
gotmax23, line 71: Why not

```suggestion
                  outputs_file.write(f'{name}={value}\n')
```
? You shouldn't need `writelines` to write one line and `\n` does the same thing as `os.linesep` (in text mode, python converts `\n` to `\r\n` if that's `os.linesep` and we're not working with Windows to begin with).

.github/workflows/ansible-release.yml
```diff
@@ -49,10 +47,39 @@ jobs:
...
+
+          try:
+              version = Version(VERSION)
+          except Exception as exc:
+              print(
+                  f'::error ::The version {VERSION!r} cannot be parsed: {exc}.'
+              )
+              sys.exit(1)
```
gotmax23, line 79: Can be condensed like this if you prefer:

```suggestion
              sys.exit(
                  f'::error ::The version {VERSION!r} cannot be parsed: {exc}.'
              )
```

.github/workflows/ansible-release.yml
```diff
@@ -84,24 +111,26 @@ jobs:
...
           ARTIFACT_URL: ${{ steps.upload-artifact.outputs.artifact-url }}
         run: |
           body="$(echo -e "${CI_COMMIT_MESSAGE}\nRelease artifacts: <${ARTIFACT_URL}>")"
+          echo -n "pr_url=" >> "$GITHUB_OUTPUT"
           gh pr create \
           --base main \
           --head "publish-${ANSIBLE_VERSION}" \
           --title "Release Ansible ${ANSIBLE_VERSION}" \
-          --body "${body}"
+          --body "${body}" >> "$GITHUB_OUTPUT"
```
gotmax23, line 133: I would use tee so the PR url is still included in the output like before

 ```suggestion
          --body "${body}" | tee -a "$GITHUB_OUTPUT"
```

File: .github/workflows/ansible-release.yml [hunk 5]
```diff
@@ -117,6 +146,16 @@ jobs:
 
     steps:
 
+    - name: Ensure that the PR was merged
+      env:
+        PR_URL: ${{ needs.build.outputs.create-pr }}
+      run: |
+        STATE=$(gh pr view "${PR_URL}" --json state --template "{{.state}}")
```
gotmax23, line 153:
```suggestion
        STATE="$(gh pr view "${PR_URL}" --json state --template "{{.state}}")"
```

gotmax23 (commented):
This looks good sans a few style nitpicks. Thanks!

.github/workflows/ansible-release.yml
felixfontein: That prints to stderr and not to stdout. Does GHA's `::error ::...` syntax also work with stderr?

.github/workflows/ansible-release.yml
```diff
@@ -84,24 +111,26 @@ jobs:
...
           ARTIFACT_URL: ${{ steps.upload-artifact.outputs.artifact-url }}
         run: |
           body="$(echo -e "${CI_COMMIT_MESSAGE}\nRelease artifacts: <${ARTIFACT_URL}>")"
+          echo -n "pr_url=" >> "$GITHUB_OUTPUT"
           gh pr create \
           --base main \
           --head "publish-${ANSIBLE_VERSION}" \
           --title "Release Ansible ${ANSIBLE_VERSION}" \
-          --body "${body}"
+         --body "${body}" | tee -a "$GITHUB_OUTPUT"
```
felixfontein, line 133:
```suggestion
          --body "${body}" | tee -a "$GITHUB_OUTPUT"
```

felixfontein: Re-add space. (I checked the original suggestion, that one didn't seem to remove this space. Maybe yet another bug in GH's suggestions?)

.github/workflows/ansible-release.yml
felixfontein: I copied this and the below from https://github.com/ansible-community/ansible-test-gh-action, it seems to be @webknjaz's preferred way of outputting things from a Python script in GHA. Maybe he can comment on it :)

.github/workflows/ansible-release.yml
gotmax23: No idea. Probably better to leave it as.

/resolve

.github/workflows/ansible-release.yml
gotmax23: Okay, fair enough

webknjaz: Yeah, I like to name the data. Otherwise, "what's is that weird `'a'` string constant being passed to a function as a positional argument?" adds some cognitive load for people that are out of context. Some day, I'll move most of that to an actual Python file and this will reduce the boilerplate too..

.github/workflows/ansible-release.yml
webknjaz: This is also copied from my code :)

I have helpers in other places where you can pass an iterable of strings and this is more genetic.

.github/workflows/ansible-release.yml
webknjaz: > Does GHA's ::error ::... syntax also work with stderr?

Yes, it works, I've done this in the past.

webknjaz:
@felixfontein one thing I'd check is whether there's enough privileges in that job to query the PR data in that publish job. You may need a `contents: read` in there (don't forget to add a code comment explaining why).

gotmax23:
> @felixfontein one thing I'd check is whether there's enough privileges in that job to query the PR data in that publish job. You may need a `contents: read` in there (don't forget to add a code comment explaining why).

I don't think it really matters if the repo is public.

.github/workflows/ansible-release.yml
```diff
@@ -49,10 +47,39 @@ jobs:
       - name: Install dependencies
         working-directory: antsibull
         run: |
+          python3 -m pip install packaging
           python3 -m pip install ansible-core
           python3 -m pip install antsibull
```
gotmax23, line 52: TODO in a separate PR: merge these into one pip install command instead of three

gotmax23 (approved):

.github/workflows/ansible-release.yml
felixfontein: I was wondering whether the separate installs were intentional for some reason, so I added a third. But yeah, let's collapse them (in a separate PR).

felixfontein:
> > @felixfontein one thing I'd check is whether there's enough privileges in that job to query the PR data in that publish job. You may need a `contents: read` in there (don't forget to add a code comment explaining why).
> 
> I don't think it really matters if the repo is public.

I *think* so too; I guess we'll see when we try this out. I'd abuse the next 10.0.0 alpha for this kind of testing :)

(I'll also try to approve the next step without merging the PR first, to see what happens. My guess is that restarting the failed second stage will work once the PR is merged.)

felixfontein:
@gotmax23 @webknjaz thanks for reviewing this!

webknjaz:
> > I don't think it really matters if the repo is public.
> 
> I _think_ so too; I guess we'll see when we try this out. I'd abuse the next 10.0.0 alpha for this kind of testing :)

@gotmax23 it does matter because setting permissions overrides the defaults and essentially revokes all other permissions that are not set explicitly. So the resulting `GITHUB_TOKEN` may end up not having enough privileges. Hence the suggestion to check it. I think I've hit a problem with this before.

felixfontein:
https://github.com/ansible-community/ansible-build-data/actions/runs/8898520146/job/24436000110
```
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
```
Fixed in 1c10087302aa09cb77679c13ac34b4cc866da162.

webknjaz:
> GH_TOKEN

`GITHUB_TOKEN` also works FYI

felixfontein:
Next try:
```
failed to run git: fatal: not a git repository (or any of the parent directories): .git
```
Fixed in acaf8bb8c5a425640b6efc694ff685895ae1d7a2.

felixfontein:
Ok, that fix was wrong, 9364168e1f889b40e2c5dffc1dc7ac4a11ece503 should do better.

felixfontein:
Now it looks good:
https://github.com/ansible-community/ansible-build-data/actions/runs/8900826549/job/24443463257
```
Error: The state of PR https://github.com/ansible-community/ansible-build-data/pull/400 must be MERGED, not OPEN
```
:tada:

PR #395, opened 2024-04-26
Release workflow: simplify pip install

felixfontein:
Ref: https://github.com/ansible-community/ansible-build-data/pull/390#discussion_r1578364641

### DIFF
--- a/.github/workflows/ansible-release.yml
+++ b/.github/workflows/ansible-release.yml
@@ -47,9 +47,7 @@ jobs:
       - name: Install dependencies
         working-directory: antsibull
         run: |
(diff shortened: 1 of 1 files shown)


webknjaz (approved):
FWIW it's not *just* a simplification — requesting all the requirements in the same install session lets pip take them into account during the dependency resolution. With 3 separate install sessions, pip could introduce dependencies incompatible with parts of earlier installations. It'd be even better to add pip-tools into the soup later ;)

felixfontein:
Thanks for reviewing this @webknjaz!

PR #396, opened 2024-04-27
Add linting workflow with yamllint; fix linting issues

felixfontein:
The indents in the release workflow were different for different parts of the file. This fixes it, and adds a yamllint step for the whole repository.

The automatically generated/updated files are ignored.


gotmax23:
I would prefer to have a nox or tox or pre-commit configuration (whichever you prefer but perhaps a good idea to standardize on nox) to run yamllint. This would make it easier to add others linters in the future and should make things easier for contributors.

felixfontein:
In that case I'd prefer `nox` over both the other tools.

felixfontein:
There's now a new `nox` session `lint`.

mariolenz (commented):
LGTM but I don't know `nox` good enough to say if this is really the best solution. So I'll only comment and not approve.

Anyway: Release early, release often! We can improve this later. If @gotmax23 approves, take this as an approval from me also.

@felixfontein I'm no expert on GH workflows but it looks like you've implemented that a new release can't be published on PyPI when the PR isn't merged yet...? If I'm right, I think that's a good idea! You know why :wink:

felixfontein:
> @felixfontein I'm no expert on GH workflows but it looks like you've implemented that a new release can't be published on PyPI when the PR isn't merged yet...? If I'm right, I think that's a good idea! You know why 😉

Nope, this PR is unrelated to that ... #390 should prevent accidental releases. I guess we can test it tomorrow with 10.0.0a2, since ansible-core 2.17.0rc1 got released today.

felixfontein:
If nobody objects, I'm going to merge this soon.

PR #419, opened 2024-06-04
Add kubevirt.core to Ansible 10 and 11

felixfontein:
Ref: https://forum.ansible.com/t/5084
Ref: https://github.com/ansible-collections/ansible-inclusion/discussions/70#discussioncomment-9650554

I would suggest to wait a few days before merging this, in case we need to do an emergency 10.0.1 release for whatever (unexpected) reason. If we haven't done one in a few days it's safe to assume the next release will be 10.1.0, which accepts new features (and thus new collections).

### DIFF
--- a/10/ansible-10.build
+++ b/10/ansible-10.build
@@ -73,6 +73,7 @@ inspur.sm: >=2.3.0,<3.0.0
 junipernetworks.junos: >=8.0.0,<9.0.0
 kaytus.ksmanage: >=1.2.0,<2.0.0
 kubernetes.core: >=3.1.0,<4.0.0
+kubevirt.core: >=1.4.0,<2.0.0
 lowlydba.sqlserver: >=2.3.0,<3.0.0
 microsoft.ad: >=1.5.0,<2.0.0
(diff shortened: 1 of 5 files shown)


mariolenz:
Just wanted to ask if this shouldn't be added to `11/` also once #420 has been merged :laughing:

felixfontein:
And indeed there was a 10.0.1 release :) Let's wait a few more days, not that we need to do yet another 10.0.2 release...

felixfontein:
The next minor releases are coming up (next week), so let's merge this ;)

felixfontein:
Hmm, maybe merging this was a mistake. kubevirt.core has a very specific kubernetes.core dependency: `>=3.1.0,<4.1.0`. First the `<4.1.0` should really **not** be there, it should be `<5.0.0`. Also kubernetes.core 5.0.0 is out (since 5 days already), so bumping that to `<6.0.0` should be done soon as well.

felixfontein:
I created https://github.com/kubevirt/kubevirt.core/issues/110.

felixfontein:
#433 re-adds this.

PR #421, opened 2024-06-04
Improve Automated Release Process documentation

mariolenz:
I've done the 10.0.0 release today and it turned out that the documentation doesn't say you need to update `_ansible_version` in the `.deps` file for this and other special releases.


felixfontein (approved):

felixfontein:
Thanks for fixing the docs, @mariolenz!

PR #422, opened 2024-06-04
Add reference to preparing next major release doc; reference automated workflow from freeze docs

felixfontein:
More docs improvements.

### DIFF
--- a/docs/automated-process.md
+++ b/docs/automated-process.md
@@ -149,3 +149,7 @@ next to the release version:
 
 2. Set `existing-branch` to the branch you pushed to the `ansible-build-data`
    repository.
+
+Finally, remember to prepare the directory for the next major Ansible release. How
+to do this is described in [Setting up for a new major
+release](new-ansible-and-freezes.md#setting-up-for-a-new-major-release).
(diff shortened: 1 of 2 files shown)


docs/automated-process.md
```diff
@@ -149,3 +149,7 @@ next to the release version:
 
 2. Set `existing-branch` to the branch you pushed to the `ansible-build-data`
    repository.
+
+Finally, remember to prepare the directory for the next major Ansible release. How
+to do this is described in [Setting up for a new major
+release](new-ansible-and-freezes.md#setting-up-for-a-new-major-release).
```
mariolenz, line 155: I'm not 100% sure about the "finally" here. When should this be done? When the new major release is available on PyPI? At least, that's how I always understood this.

Maybe we should make this clearer by saying "When the new major release has been done and is available on PyPI, remember to prepare..." or something similar.

felixfontein: Basically it can be done once the PR has been merged, i.e. as soon as the `X/ansible-X.0.0.deps` file is there. I'll update the formulation later.

felixfontein: Updated in af5d0338e652235e72ce3e6ad53befce531aa19f.

felixfontein:

PR #423, opened 2024-06-04
Detail about playbook conditionals from changelog

oraNod:
This change manually includes details from the 2.17.0 changelog that were not included in the porting guide.

Relates to https://github.com/ansible/ansible-documentation/pull/1523

### DIFF
--- a/10/porting_guide_10.rst
+++ b/10/porting_guide_10.rst
@@ -18,7 +18,27 @@ We suggest you read this page along with the `Ansible 10 Changelog <https://gith
 Playbook
 ========
 
-No notable changes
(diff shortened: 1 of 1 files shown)


felixfontein (changes requested):
This has to be fixed in ansibe/ansible-documentation, not here.

felixfontein (approved):
Ah, I just saw it's already merged there.

samccann:
Just curious - what happens to this manual patch when 10.1.0 is released? Will it still show up?

felixfontein:
It should stay, since by then this part is replaced by the part in ansible-documentation (either stable-2.17 branch or devel, I would have to check the code), and it's included in there as well.

felixfontein:
@oraNod thanks!

PR #438, opened 2024-06-21
Sort collection-meta.yaml

mariolenz:
As announced in the last community meeting, I would like to sort collection-meta.yaml alphabetically.

I've sorted this with `grep '^  [^ ]' collection-meta.yaml | sort` and there might be differences between locales, but I think it's an improvement to the current situation.

_edit_: +455 / -455 lines, so it looks like I didn't drop any information accidentally :-)

### DIFF
--- a/10/collection-meta.yaml
+++ b/10/collection-meta.yaml
@@ -1,156 +1,17 @@
 ---
 collections:
-  ansible.utils:
-    maintainers:
-      - cidrblock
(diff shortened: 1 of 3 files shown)


samccann (commented):
LGTM. Is it worth adding a note at the top to say 'keep this in alphabetical order' :-)

felixfontein (approved):

PR #431, opened 2024-06-15
Add ieisystem.inmanage to Ansible 10 and 11

mariolenz:
Ref: https://forum.ansible.com/t/6209
Ref: ansible-collections/ansible-inclusion#66

PR #439, opened 2024-06-30
Add vmware.vmware  to Ansible 10 and 11

mariolenz:
Ref: https://forum.ansible.com/t/6768
Ref: ansible-collections/ansible-inclusion#75

### DIFF
--- a/10/collection-meta.yaml
+++ b/10/collection-meta.yaml
@@ -1,3 +1,4 @@
+# Please keep this in alphabetical order
 ---
 collections:
   amazon.aws:
@@ -287,6 +288,8 @@ collections:
       - rndmh3ro
       - schurzi
     repository: https://github.com/T-Systems-MMS/ansible-collection-icinga-director
+  vmware.vmware:
+    repository: https://github.com/ansible-collections/vmware.vmware
   vmware.vmware_rest:
     maintainers:
       # ansible-collections/cloud
--- a/10/ansible-10.build
+++ b/10/ansible-10.build
@@ -93,6 +93,7 @@ splunk.es: >=3.0.0,<4.0.0
 t_systems_mms.icinga_director: >=2.0.0,<3.0.0
 telekom_mms.icinga_director: >=2.1.0,<3.0.0
 theforeman.foreman: >=4.0.0,<5.0.0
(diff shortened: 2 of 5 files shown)


felixfontein:
Out of curiosity: why not add it to Ansible 9 as well?

mariolenz:
Maybe I got this wrong, but I thought Ansible 9 is kind of an LTS release where we don't add new collections. We didn't add `ieisystem.inmanage` there, either: #431 

And your PR to add `kubevirt.core` is also about 10 and 11, not 9: #419 

I think `vmware.vmware` should work in Ansible 9 because they claim to work with ansible-core >= 2.15.0. But if we add it to 9, we should also talk about adding `ieisystem.inmanage` and `kubevirt.core`, too.

I'm not against adding this collection to the Ansible Community Package 9, I just thought we don't want to do this :shrug: 

Correct me I've I'm wrong. But, as I've said, we also should have another look at `ieisystem.inmanage` and `kubevirt.core`.

felixfontein:
It's not really LTS, since it accepts new features in collections. So why not also add new collections?

For ieisystem.inmanage that was likely an oversight, and we should fix that.

For kubevirt.core that was intentional, since it depends on kubernetes.core >= 3.1.0, while Ansible 9 contains kubernetes.core 2.x.y. So it's not possible to add it to Ansible 9. (See also https://forum.ansible.com/t/5084/8 :) )

10/collection-meta.yaml
```diff
@@ -287,6 +288,8 @@ collections:
       - rndmh3ro
       - schurzi
     repository: https://github.com/T-Systems-MMS/ansible-collection-icinga-director
+  vmware.vmware:
+    repository: https://github.com/ansible-collections/vmware.vmware
```
samccann, line 292: Can we hold this PR until we get at least one maintainer to list here?

felixfontein: The inclusion request comes with a list of GH handles:

`GitHub Handles: @machacekondra @mikemorency @bardielle @mariolenz`

These should be added as maintainers.

felixfontein (approved):

samccann (approved):

mariolenz:
@felixfontein @samccann Thanks for reviewing!

PR #451, opened 2024-08-27
Order collections lexicographically

felixfontein:
Subset of #450 which doesn't have new metadata, but simply reorders the collections. Will rebase that after this is merged.


mariolenz:
@felixfontein Out of curiosity, how did you order the collections? Did you do it manually, or did you use a tool?

In #438 I've simply used `sort`. How did you do it?

felixfontein:
@mariolenz I did it manually, based on the feedback from the linter (https://github.com/ansible-community/antsibull/pull/617). The linter uses Python's `sorted()`.

PR #456, opened 2024-09-13
Deprecate frr.frr

mariolenz:
This is just meant as an example how to deprecate a collection in the Ansible Community Package.


felixfontein:
Please note that this will no be a good example soon anymore, once #450 has been merged (and potentially some follow-up PRs for antisbull which take care of changelog entries).

PR #458, opened 2024-09-20
t_systems_mms.icinga_director should have been remove from Ansible 11

felixfontein:
The collection was renamed, and replaced by deprecated redirects in Ansible 9. Fix metadata, remove old collection from Ansible 11.

### DIFF
--- a/10/changelog.yaml
+++ b/10/changelog.yaml
@@ -134,3 +134,11 @@ releases:
 
         `Porting Guide <https://docs.ansible.com/ansible/devel/porting_guides.html>`_'
     release_date: '2024-09-10'
+  10.5.0:
+    changes:
+      deprecated_features:
+      - The collection ``t_systems_mms.icinga_director`` has been renamed to ``telekom_mms.icinga_director``.
+        For now both collections are included in Ansible. The content in ``t_systems_mms.icinga_director``
+        has been replaced with deprecated redirects to the new collection in Ansible
+        9.0.0, and these redirects will be removed from Ansible 11. Please update
+        your FQCNs for ``t_systems_mms.icinga_director``.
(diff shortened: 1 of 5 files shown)


felixfontein:
Needs https://github.com/ansible-community/antsibull-core/pull/172 to pass CI.

10/changelog.yaml
```diff
@@ -134,3 +134,11 @@ releases:
...
         `Porting Guide <https://docs.ansible.com/ansible/devel/porting_guides.html>`_'
     release_date: '2024-09-10'
+  10.5.0:
+    changes:
+      deprecated_features:
+      - The collection ``t_systems_mms.icinga_director`` has been renamed to ``telekom_mms.icinga_director``.
+        For now both collections are included in Ansible. The content in ``t_systems_mms.icinga_director``
+        has been replaced with deprecated redirects to the new collection in Ansible
+        9.0.0, and these redirects will be removed from Ansible 11. Please update
+        your FQCNs for ``t_systems_mms.icinga_director``.
```
gotmax23, line 144: This is an optional suggestion, but I think it makes much clearer what users need to do to adapt to the deprecation, even if we already have both names in the first sentence.

```suggestion
        your FQCNs from ``t_systems_mms.icinga_director`` to ``telekom_mms.icinga_director``.
```

gotmax23 (approved):

10/collection-meta.yaml
```diff
@@ -309,6 +309,12 @@ collections:
       - rndmh3ro
       - schurzi
     repository: https://github.com/T-Systems-MMS/ansible-collection-icinga-director
+    removal:
+      major_version: 11
+      reason: renamed
+      new_name: telekom_mms.icinga_director
+      announce_version: 10.5.0
```
gotmax23, line 316: Why are announce versions different for each major version? Why isn't is also `9.0.0a1` here? Isn't that when it was first announced?

felixfontein: Basically this says when the announcement appears in the changelog of this major Ansible version. We need this information to be able to render the changelog.

felixfontein:
@gotmax23 thanks for reviewing!

gotmax23:
Thanks!

10/collection-meta.yaml
gotmax23: Got it

PR #470, opened 2024-10-08
Ansible 11: Restrict ansible.posix to < 1.6.0

mariolenz:
1.6.0 has a breaking change: the skippy plugin has been removed. This is a semantic versioning violation.

Note to self: Have a look at the other constraints!

### DIFF
--- a/11/ansible-11.constraints
+++ b/11/ansible-11.constraints
@@ -1,0 +1,2 @@
+# ansible.posix 1.6.0 has a breaking change
+ansible.posix: <1.6.0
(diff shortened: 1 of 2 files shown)


11/ansible-11.constraints
```diff
@@ -2,3 +2,5 @@
 cloud.common: <4.0.0
 # cisco.dnac 6.16.0, cisco.ise 2.9.2, and cisco.meraki 2.18.1 need ansible.utils < 5.0.0
 ansible.utils: <5.0.0
```
felixfontein, line 4: Note to self: we have to look at the other constraints, whether we can now remove them. If we cannot, we might have to remind folks or remove some collections. (https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_requirements.html#collection-dependencies)

felixfontein: They actually were no longer necessary, I removed them in #471. This unfortunately means I introduced a conflict, sorry for that :)

felixfontein:
Since we reverted this for the other versions, let's close this. Thanks for creating the PR!

Issue #476, opened 2024-10-09
Remove google.cloud from Ansible 12


gotmax23:
https://forum.ansible.com/t/8609

rpavlik:
It appears this is fixed? Kind of weird a vote for removal was somehow easier than just fixing the doc errors. https://github.com/ansible-collections/google.cloud/pull/655

felixfontein:
It's not just about "docs errors". Also please use https://forum.ansible.com/t/8609 to discuss the removal, this issue is just for tracking the actual removal.

PR #477, opened 2024-10-09
Schedule google.cloud for removal from Ansible 12

gotmax23:
Relates: https://forum.ansible.com/t/8609
Relates: https://github.com/ansible-community/ansible-build-data/issues/476

### DIFF
--- a/11/collection-meta.yaml
+++ b/11/collection-meta.yaml
@@ -209,6 +209,14 @@ collections:
     repository: https://github.com/fortinet-ansible-dev/ansible-galaxy-fortios-collection
   google.cloud:
     repository: https://github.com/ansible-collections/google.cloud
+    removal:
+      major_version: 12
+      announce_version: 11.0.0a2
+      reason: guidelines-violation
+      discussion: https://forum.ansible.com/t/8609
+      reason_text: >
+        The collection has
+        L(unresolved sanity test failures, https://github.com/ansible-collections/google.cloud/issues/613).
   grafana.grafana:
     maintainers:
       - ishanjainn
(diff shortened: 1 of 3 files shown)


felixfontein (approved):
Should we wait with merging this until https://forum.ansible.com/t/9684/10 happened? Assuming that meeting happens before the next Ansible releases (that aren't pre-releases).

mariolenz (approved):
> Should we wait with merging this until https://forum.ansible.com/t/9684/10 happened? Assuming that meeting happens before the next Ansible releases (that aren't pre-releases).

The vote about  removing the collection has been accepted, so I think we should do it. If we make a decision, I think we should follow it!

gotmax23:
> The vote about removing the collection has been accepted, so I think we should do it. If we make a decision, I think we should follow it!

Yeah, I agree, unless there's some strong reason not to, in which case that should be discussed in a community topic.

/cc @gundalow

felixfontein:
In that case, feel free to merge this.

11/collection-meta.yaml
```diff
@@ -209,6 +209,14 @@ collections:
     repository: https://github.com/fortinet-ansible-dev/ansible-galaxy-fortios-collection
   google.cloud:
     repository: https://github.com/ansible-collections/google.cloud
+    removal:
+      major_version: 12
+      announce_version: 11.0.0a2
```
felixfontein, line 214: Assuming there is no further alpha release, this needs to be updated to:
```suggestion
      announce_version: 11.0.0b1
```

felixfontein:
@gotmax23 @mariolenz thanks!

PR #469, opened 2024-10-08
Restrict ansible.posix to < 1.6.0

felixfontein:
1.6.0 has a breaking change: the skippy plugin has been removed. This is a semantic versioning violation.

PR #478, opened 2024-10-11
ansible.posix reverted the breaking change

mariolenz:
It looks like [ansible.posix 1.6.1](https://github.com/ansible-collections/ansible.posix/releases/tag/1.6.1) reverted the breaking change that has been introduced in 1.6.0. We should undo / revert #469.

Ref: https://github.com/ansible-collections/ansible.posix/commit/cd43bd10bbf4401582ef2305dfc74286b9da05a2
Ref: ansible-collections/ansible.posix#574

### DIFF
--- a/10/ansible-10.constraints
+++ b/10/ansible-10.constraints
@@ -1,2 +1,0 @@
-# ansible.posix 1.6.0 has a breaking change
-ansible.posix: <1.6.0
--- a/9/ansible-9.constraints
+++ b/9/ansible-9.constraints
@@ -1,2 +1,0 @@
-# ansible.posix 1.6.0 has a breaking change
-ansible.posix: <1.6.0


felixfontein:
We should probably add a changelog entry that ends up in the porting guide that says that the breaking change has been reverted. Otherwise users only see a breaking change in the porting guide but don't see that it is no longer in there, due to both the 1.6.0 and 1.6.1 changelogs being included.

felixfontein:
It's probably better to have a mechanism to hide specific changelog entries from the combined changelog. I might work on that during the next days.

mariolenz:
Something like `ignore-changelog-entries.yml`:

```
# Ignore removal because it has been undone
ansible.posix:
  1.6.0:
    removed_features:
    - skippy - Remove skippy pluglin as it is no longer supported(https://github.com/ansible-collections/ansible.posix/issues/350).
```

or something similar? Yes, I can see how this could be helpful in cases like this.

felixfontein:
I'd probably try to integrate this into `changelog.yaml` to avoid having yet another file to handle...

felixfontein (approved):

felixfontein:
@mariolenz thanks!

PR #485, opened 2024-10-19
Bump Python versions

felixfontein:
Use Python 3.12 instead of 3.11 for building Ansible (all versions of Ansible currently built should support 3.12).

Use Python 3.13 instead of 3.12 for linting.


mariolenz (approved):
LGTM

felixfontein:

PR #489, opened 2024-11-03
[WIP] Re-add netapp.storagegrid

mariolenz:
This depends on https://forum.ansible.com/t/2811/24

Related:
#373
#434
#488

PR #495, opened 2024-11-06
Re-add netapp.storagegrid to Ansible 11

felixfontein:
We discussed the vote and [action on it on Monday](https://forum.ansible.com/t/2811/28) at today's community meeting. My interpretation of the process was definitely wrong, in case of missed quorum we should have re-added the collection. (We need to improve the formulation (right now it's quite convoluted) and also to clarify some other aspects.) We decided to do another beta release, 11.0.0b1, today with netapp.storagegrid re-added.

This is a resurrection of #489.


mariolenz (approved):
LGTM

felixfontein:

PR #501, opened 2024-11-16
Remove ibm.spectrum_virtualize from Ansible 12

### LINKED ISSUE #306: Rename included ibm.spectrum_virtualize to ibm.storage_virtualize
Relates to https://github.com/ansible-collections/ansible-inclusion/discussions/65

The renaming process is described in [README](https://github.com/ansible-community/ansible-build-data/#renaming-a-collection)

For simplicity, assume that the next minor Ansible release is X.Y.0, and that collection foo.bar with latest release a.b.c has been renamed to baz.bam with latest release A.B.C.
- [x] 1. baz.bam A.B.C must be compatible to foo.bar a.b.c up to renaming plugins. No options must be renamed without backwards compatible aliases, and no defaults or semantics changed.
- [x] 2. baz.bam A.B.C can be added to Ansible X.Y.0.
- [x] 3. A deprecation warning is added to Ansible X.Y.0's changelog (deprecated_features) that foo.bar has been renamed to baz.bam, that Ansible (X+1).0.0 will start having deprecated redirects from foo.bar to baz.bam, and that foo.bar will be removed from a later major release of Ansible.
- [ ] 4. A new release foo.bar (a+1).0.0 is made which contains no more content, but only deprecated redirects to baz.bam. Ideally it will have a dependency on baz.bam so that users that install foo.bar will have working deprecated redirects.
- [ ] 5. Ansible (X+1).0.0 contains both foo.bar (a+1).0.0, and either a baz.bam A.B'.C' release or a later major release that is still compatible with foo.bar a.b.c as specified in 1.
- [ ] 6. foo.bar will be dropped from Ansible (X+2).0.0 (needs to be announced in its changelog as removed_features).

felixfontein:
The collection hasn't been replaced by redirects *again*, so let's just remove it.

Closes #306.

See also https://forum.ansible.com/t/8119

### DIFF
--- a/11/collection-meta.yaml
+++ b/11/collection-meta.yaml
@@ -235,8 +235,9 @@ collections:
       - Shilpi-J
     repository: https://github.com/ansible-collections/ibm.spectrum_virtualize
     removal:
-      major_version: TBD
+      major_version: 12
       reason: renamed
+      announce_version: 11.1.0
       new_name: ibm.storage_virtualize
   ibm.storage_virtualize:
     maintainers:
--- a/10/collection-meta.yaml
+++ b/10/collection-meta.yaml
@@ -242,8 +242,9 @@ collections:
       - Shilpi-J
     repository: https://github.com/ansible-collections/ibm.spectrum_virtualize
     removal:
-      major_version: TBD
+      major_version: 12
       reason: renamed
+      announce_version: 10.7.0
       new_name: ibm.storage_virtualize
   ibm.storage_virtualize:
     maintainers:
--- a/12/ansible.in
+++ b/12/ansible.in
@@ -58,7 +58,6 @@ fortinet.fortios
 grafana.grafana
 hetzner.hcloud
 ibm.qradar
-ibm.spectrum_virtualize
 ibm.storage_virtualize
 ieisystem.inmanage
 infinidat.infinibox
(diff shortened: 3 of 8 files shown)


11/collection-meta.yaml
```diff
@@ -235,8 +235,9 @@ collections:
       - Shilpi-J
     repository: https://github.com/ansible-collections/ibm.spectrum_virtualize
     removal:
-      major_version: TBD
+      major_version: 12
       reason: renamed
+      announce_version: 11.0.0
```
mariolenz, line 240: This means 11.0.0 wouldn't be _exactly_ the same as 11.0.0rc1. I'm not sure if this is really a problem. On the other hand: Why not wait until 11.1.0? It's only 2 weeks later. And since we want to release 11.0.0 in two days, this is on very short notice

felixfontein: Well, 11.0.0 is NEVER exacly like 11.0.0rc1, since 1) it has a different version, and 2) the changelog and porting guide look very different. For example compare the changelog of 10.0.0rc1 (https://github.com/ansible-community/ansible-build-data/blob/10.0.0rc1/10/CHANGELOG-v10.md#v10-0-0) with the one of 10.0.0 (https://github.com/ansible-community/ansible-build-data/blob/10.0.0/10/CHANGELOG-v10.md). So having an additional changelog entry announcing the removal isn't a big difference as opposed to all the other changes.

felixfontein: It depends, either this gets enough approval and we can merge it before the 11.0.0 release, or I'll bump the version here and we merge it for 11.1.0. Or later, if there aren't enough approvals until then either.

mariolenz (approved):
LGTM

felixfontein:
I've updated the removal version to 11.1.0.

Andersson007 (approved):

PR #514, opened 2025-01-19
Updated changelogs and porting guides with notes that removed collections can still be installed manually

felixfontein:
Updated with the changes in https://github.com/ansible-community/antsibull-build/pull/647.


mariolenz:
I don't see this for `community.network` in ansible-community/ansible-build-data#514. It's not a good idea to use an officially unmaintained collection, but why not add the information that the collection can still be installed via galaxy?

I'm not saying that we should add this. It's just something that I've stumbled upon and wanted to discuss.

felixfontein:
That's because community.network has been deprecated (I kind of remember there was talk of even archiving the repository) and is not just unmaintained. While you can still install it manually, I'm not sure we should advertise that. (I used the same conditions as in the case when collections are actually removed - there we already had this sentence there, but not for renamed or deprecated collections.)

If someone wants to show it also for deprecated collections, I'd like to see a good reason for it :)

mariolenz:
> If someone wants to show it also for deprecated collections, I'd like to see a good reason for it :)

I don't have a good reason. Just wanted to mention it ;-)

samccann (approved):

PR #520, opened 2025-02-24
Un-restrict vmware.vmware

mariolenz:
It looks like the collection is correctly tagged now, so lets' undo #519.


mariolenz:
@felixfontein Looks like the collection is tagged correctly now. Should we remove the restriction?

If you agree, should we also close ansible-collections/vmware.vmware#124?

BTW thanks for spotting this!

felixfontein:
@mariolenz thanks!

PR #522, opened 2025-03-24
Restrict ibm.storage_virtualize to <2.7.2

felixfontein:
Ref: https://github.com/ansible-collections/ibm.storage_virtualize/issues/57
Ref: https://github.com/ansible-community/ansible-build-data/issues/223#issuecomment-2749183864

### DIFF
--- a/11/ansible-11.constraints
+++ b/11/ansible-11.constraints
@@ -1,0 +1,2 @@
+# ibm.storage_virtualize 2.7.2 is not tagged (https://github.com/ansible-collections/ibm.storage_virtualize/issues/57)
+ibm.storage_virtualize: <2.7.2
--- a/12/ansible-12.constraints
+++ b/12/ansible-12.constraints
@@ -1,0 +1,2 @@
+# ibm.storage_virtualize 2.7.2 is not tagged (https://github.com/ansible-collections/ibm.storage_virtualize/issues/57)
+ibm.storage_virtualize: <2.7.2


11/ansible-11.constraints
```diff
@@ -0,0 +1,2 @@
+# ibm.storage_virtualize 2.7.2 is not tagged (https://github.com/ansible-collections/ibm.storage_virtualize/issues/57)
+ibm.storage_virtualize: <2.7.2
```
mariolenz, line 2:
```suggestion
# ibm.storage_virtualize 2.6.0+ is not tagged (https://github.com/ansible-collections/ibm.storage_virtualize/issues/57)
ibm.storage_virtualize: <=2.6.0
```

It looks like the last tag is 2.6.0. So I wonder if we shouldn't be more strict here and disallow anything later than 2.6.0.

12/ansible-12.constraints
```diff
@@ -0,0 +1,2 @@
+# ibm.storage_virtualize 2.7.2 is not tagged (https://github.com/ansible-collections/ibm.storage_virtualize/issues/57)
+ibm.storage_virtualize: <2.7.2
```
mariolenz, line 2:
```suggestion
# ibm.storage_virtualize 2.6.0+ is not tagged (https://github.com/ansible-collections/ibm.storage_virtualize/issues/57)
ibm.storage_virtualize: <=2.6.0
```

It looks like the last tag is 2.6.0. So I wonder if we shouldn't be more strict here and disallow anything later than 2.6.0.

mariolenz (changes requested):
It looks like 2.7.0 and 2.7.1 aren't tagged, either.

Or am I wrong there?

11/ansible-11.constraints
felixfontein: 2.7.0 and 2.7.1 don't exist on Galaxy, either. So technically it's not a problem that they're not tagged, since from our POV they don't exist...

12/ansible-12.constraints
felixfontein: Releases that don't exist don't count ;)

11/ansible-11.constraints
mariolenz: Oh, you're right. I didn't check this on galaxy. I just assumed that if there's a 2.7.2, there are also 2.7.0 and 2.7.1. Silly me!

12/ansible-12.constraints
mariolenz: Oh, you're right. I didn't check this on galaxy. I just assumed that if there's a 2.7.2, there are also 2.7.0 and 2.7.1. Silly me!

felixfontein:
Since there's still no tag, I'm going to merge this.

PR #553, opened 2025-06-03
Modify Ansible 12.0.0a4 to contain netapp.storagegrid 21.14.0 instead of 21.15.0

felixfontein:
Ref: https://forum.ansible.com/t/43132


felixfontein:
@gotmax23 @mariolenz thanks!

PR #565, opened 2025-07-03
Adds workflow to upload to test.pypi

anweshadas:
This is a workflow to help the release managers to get accustomed with the release process.
This gives us the ability to run tests without messing the real PyPI.


felixfontein:
> Updated both the PR description to mark that this has to run on maintainer's fork

Would be great to make it fail early if it's run in this repo instead of a fork.

PR #570, opened 2025-07-13
Run Ansible test builds daily

felixfontein:
Right now they run once per week (on Thursday), which can be problematic if a bad release happens afterwards and before the next release (which is usually on Tuesday). I think it would be better to run the test builds daily so we find out early about problems.


felixfontein:
@mariolenz thanks for reviewing!

PR #579, opened 2025-08-04
Add ravendb.ravendb to Ansible 12

mariolenz:
Ref: https://forum.ansible.com/t/44150
Ref: ansible-collections/ansible-inclusion#80

Opening as a draft, because the time to merge depends on what we decide in https://forum.ansible.com/t/44186.

### DIFF
--- a/11/ansible-11.build
+++ b/11/ansible-11.build
@@ -89,6 +89,7 @@ openstack.cloud: >=2.2.0,<3.0.0
 ovirt.ovirt: >=3.2.0,<4.0.0
 purestorage.flasharray: >=1.31.0,<2.0.0
 purestorage.flashblade: >=1.19.0,<2.0.0
+ravendb.ravendb: >=1.0.2,<2.0.0
 sensu.sensu_go: >=1.14.0,<2.0.0
 splunk.es: >=4.0.0,<5.0.0
 telekom_mms.icinga_director: >=2.2.0,<3.0.0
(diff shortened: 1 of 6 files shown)


gotmax23 (changes requested):
Blocking until the issues with CI testing raised in https://github.com/ansible-collections/ansible-inclusion/discussions/80 are addressed

gotmax23:
I would prefer to wait until 12.1.0, since the main reason for lifting the feature freeze as far as I understood was because of the network collections and to avoid holding back bugfixes for collections that are already part of the ansible package. But if others feel strongly that this should be included in the next pre-release since we already paused the feature freeze, that's fine with me.

felixfontein:
I'm fine with both (waiting and merging now). Waiting would mean that we wait at least two more weeks before merging this (assuming we do 12.0.0b4 this week, 12.0.0rc1 next week, and 12.0.0 in two weeks). For Ansible 11, this doesn't make a difference, since the 11.10.0 release is in three weeks anyway. But if 12.0.0 gets delayed by more than two weeks compared to the schedule I sketched above, this would have to wait for 11.11.0 (unless we split it up into two PRs).

@mariolenz @anweshadas what do you think?

felixfontein:
Then my suggestion would be to wait until 12.0.0 is out before merging, unless 11.10.0 happens earlier, in which case we'd split this PR up into one for Ansible 11 and one for Ansible 12, so that ravendb.ravendb will definitely be included in Ansible 11.10.0, the next Ansible 11 release. Is that OK for everyone?

felixfontein:
I created #592 from this PR that only contains the Ansible 11 changes.

mariolenz:
I've removed the Ansible 11 changes from this PR.

felixfontein (approved):

mariolenz:
@gotmax23 Now that 12.0.0 has been released, do you think we should merge this and add `ravendb.ravendb`?

felixfontein:
Maybe let's wait one more week, in case we want to do a 12.0.1 release.

PR #593, opened 2025-08-27
Remove cloud.common from Ansible 12

felixfontein:
Ref: https://forum.ansible.com/t/41507/24


felixfontein:
vmware.vmware_rest 4.9.0 is out, I've re-triggered CI. It should hopefully pass.

felixfontein:
Uhm, CI just ran?

mariolenz (approved):

mariolenz:
Maybe we should change `vmware.vmware_rest: >=4.8.0,<5.0.0` to `vmware.vmware_rest: >=4.9.0,<5.0.0` in `12/ansible-12.build` while we're at it.

felixfontein:
Since @gotmax23 wrote on Matrix last week that removing cloud.common is a consequence from our rules, I think we can count that as an approval. So let's merge it now instead of waiting longer.

felixfontein:
@mariolenz @mikemorency thanks for your work in making this possible (and reviewing this to @mariolenz :) )!

PR #601, opened 2025-09-09
Add 13/ directory

mariolenz:
Now that 12.0.0 is out, I think we should create the directory for the next major version.

This is just quick'n'dirty without checking if we wanted to remove any collections from 13. We can do this later IMHO.


mariolenz:
@felixfontein I think I've fixed one problem (with ibm.qradar) but the linter still has a problem with google.cloud. I'm not sure what to do there... should I just remove the complete `removal` section? I guess this will fix the sanity test, but I'm not sure if this is the right thing to do.

PR #604, opened 2025-09-17
Adds documentation about Release Management Mentorship program 

anweshadas:
This adds the definition, expectation  initial plan and steps for  Release Management Mentorship program .  Also shares the resources to be shared during the program. 
This is to be treated like a live document, which evolves as the program continues.

### DIFF
--- a/docs/release-managers-mentorship/training.md
+++ b/docs/release-managers-mentorship/training.md
@@ -1,0 +1,83 @@
+
+
+Mentorship Program for New Ansible Release Managers!
+
+Under this initiative, we’ll mentor and train new community members to become Ansible Release Managers.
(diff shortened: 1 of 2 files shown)


abhay-dandge:
**@anweshadas** I noticed a small issue in this commit 🙂

In `training_resources.md`, the following two entries point to the same YouTube video (`https://www.youtube.com/watch?v=V9xNuq3B69A`):

- *Using Trusted Publishing to Ansible release*  
- *Collections signal major shift in Ansible ecosystem*  

Since both links are identical, one of them looks like a duplicate. Could you please double-check and update/remove the duplicate entry?  

Thanks!

PR #664, opened 2026-03-31
Remove deprecated cisco.dnac from Ansible 14

mariolenz:
Remove [deprecated](https://github.com/cisco-en-programmability/dnacenter-ansible/pull/324) collection `cisco.dnac` from Ansible 14.

Ref: https://forum.ansible.com/t/45609

Like #660 but with (I hope correctly) updated versions.

### DIFF
--- a/14/collection-meta.yaml
+++ b/14/collection-meta.yaml
@@ -36,11 +36,6 @@ collections:
     collection-directory: "./chocolatey"
   cisco.aci:
     repository: https://github.com/CiscoDevNet/ansible-aci
-  cisco.dnac:
-    maintainers:
-      - racampos
-      - wastorga
-    repository: https://github.com/cisco-en-programmability/dnacenter-ansible
   cisco.intersight:
     repository: https://github.com/CiscoDevNet/intersight-ansible
   cisco.ios:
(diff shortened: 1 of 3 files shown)


14/collection-meta.yaml
```diff
@@ -353,6 +348,16 @@ removed_collections:
       reason: deprecated
       discussion: https://forum.ansible.com/t/38960
       announce_version: 11.2.0
+  cisco.dnac:
+    maintainers:
+      - racampos
+      - wastorga
+    repository: https://github.com/cisco-en-programmability/dnacenter-ansible
+    removal:
+      version: 14.0.0a2
```
gotmax23: Shouldn't this be

```suggestion
      version: 14.0.0a1
```
?

mariolenz: According to the [roadmap](https://docs.ansible.com/projects/ansible/devel/roadmap/COLLECTIONS_14.html), I think we'll release 14.0.0a1 before 13.6.0. So I thought a2 would be better.

@felixfontein What do you think, a1 or a2?

felixfontein: We remove it from 14.0.0a1, so it should say 14.0.0a1 even if the announcement in the 13.x.y changelog and porting guide only happens after the removal.

gotmax23 (approved):

mariolenz:
Thanks for reviewing @gotmax23 and @felixfontein!

felixfontein:
@mariolenz thanks for doing the actual work! :)

PR #673, opened 2026-04-29
Restrict cyberark.pas to < 1.0.40

felixfontein:
1.0.40 hasn't been tagged.

Ref: #223


gotmax23:
See https://github.com/cyberark/ansible-security-automation-collection/issues/90.

Issue #676, opened 2026-05-06
Rename community.mysql to ansible.mysql: post actions


Andersson007:
Hi everyone,
1. ansible.mysq has been already added to the package

I've just also done:
1. Deprecated all modules in community.mysql, release 4.3.0 https://github.com/ansible-collections/community.mysql/pull/1 
2. Removed content, added redirects, released 5.0.0 https://github.com/ansible-collections/community.mysql/pull/2 
3. Both the releases are on Galaxy https://galaxy.ansible.com/ui/repo/published/community/mysql/

I guess you need to do something here now?

felixfontein:
I think we should restrict Ansible 13 to not include community.mysql 4.3.0, as adding the replacement only in the same version when a deprecation is added is very unfriendly to users.

Please note that the community.mysql repository is missing tags for both 4.3.0 and 5.0.0. This currently makes all builds in this repository fail.

Andersson007:
> I think we should restrict Ansible 13 to not include community.mysql 4.3.0, as adding the replacement only in the same version when a deprecation is added is very unfriendly to users.

@felixfontein please do whatever is needed to make the user experience friendly

> Please note that the community.mysql repository is missing tags for both 4.3.0 and 5.0.0. This currently makes all builds in this repository fail.

oh, i though the builds here take tarballs from Galaxy, sorry. I've just added the tags, PTAL
thank you

felixfontein:
Today's "nightlies" (that ran after your post) passed. Thanks for adding the tags!

> oh, i though the builds here take tarballs from Galaxy, sorry.

It does use the tarballs, but it also verifies that the tags are there, so we can verify that the tarball's contents match the tag's contents. Having tags set for releases is a requirement: https://docs.ansible.com/projects/ansible/devel/community/collection_contributors/collection_requirements.html#repository-management

felixfontein:
I created #677 to restrict community.mysql for Ansible 13, feel free to comment.

gotmax23:
I think this is all done now?

mariolenz:
Yes, I think you're right and I'll close the issue.

@Andersson007 @felixfontein @gotmax23 Feel free to re-open it if you think I've been wrong there!

PR #677, opened 2026-05-11
Restrict community.mysql to < 4.3.0 for Ansible 13

felixfontein:
Ref: https://github.com/ansible-community/ansible-build-data/issues/676#issuecomment-4409186591

### DIFF
--- a/13/ansible-13.constraints
+++ b/13/ansible-13.constraints
@@ -12,3 +12,6 @@ junipernetworks.junos: <11.1.0
 cisco.dnac: <6.48.1
 # cyberark.pas 1.0.40 is not tagged in https://github.com/cyberark/ansible-security-automation-collection
 cyberark.pas: <1.0.40
+# community.mysql 4.3.0 deprecates all modules and suggests users to use ansible.mysql.
+# Since ansible.mysql was just added to Ansible, let's wait with this.
+community.mysql: <4.3.0
--- a/13/ansible-13.constraints
+++ b/13/ansible-13.constraints
@@ -13,5 +13,5 @@ cisco.dnac: <6.48.1
 # cyberark.pas 1.0.40 is not tagged in https://github.com/cyberark/ansible-security-automation-collection
 cyberark.pas: <1.0.40
 # community.mysql 4.3.0 deprecates all modules and suggests users to use ansible.mysql.
-# Since ansible.mysql was just added to Ansible, let's wait with this.
+# Since ansible.mysql was just added to Ansible, let's wait with this to avoid runtime deprecation warnings.
 community.mysql: <4.3.0


13/ansible-13.constraints
```diff
@@ -12,3 +12,6 @@ junipernetworks.junos: <11.1.0
 cisco.dnac: <6.48.1
 # cyberark.pas 1.0.40 is not tagged in https://github.com/cyberark/ansible-security-automation-collection
 cyberark.pas: <1.0.40
+# community.mysql 4.3.0 deprecates all modules and suggests users to use ansible.mysql.
+# Since ansible.mysql was just added to Ansible, let's wait with this.
```
mariolenz:
```suggestion
# Since ansible.mysql was just added to Ansible, let's wait with this until Ansible 13.8.0.
```

As far as I understand this is just a temporary restriction until we've released the first Ansible Community Package that contains `ansible.mysql`. This should be Ansible 13.7.0. Maybe we should make this clear in the comment.

Or do you want to wait longer?

felixfontein: I would keep it there indefinitely, so that Ansible 13 never has the deprecations. That makes Ansible 13 (from the next release on) the version that has both ansible.mysql and community.mysql without any deprecation; in Ansible 14 the community.mysql content will be replaced by redirects to ansible.mysql.

mariolenz: Are you sure? I should say that it's a good thing for users to see those deprecation warnings as early as possible.

I'm fine to keep community.mysql < 4.3.0 In Ansible 13, but I don't quite see the point to hide the fact that the collection is deprecated from users.

Anyway, I think Ansible 13.7.0 is due tomorrow. So let's merge as-is and discuss unrestricting community.mysql or not in 13.8.0 later.

gotmax23: I do think we should add a deprecation entry to the Ansible 13 changelog announcing the removal of community.mysql in favor of ansible.mysql in Ansible 15 as soon as possible (https://docs.ansible.com/projects/ansible-build-data/policies/#announce-removal-of-a-collection-deprecation). I assume @felixfontein's concerns were about the runtime deprecation warnings added to community.mysql in https://github.com/ansible-collections/community.mysql/commit/f88a7d14edd4a060d52b08c707ff6a629252c276 which are more disruptive to users.

gotmax23: > I do think we should add a deprecation entry to the Ansible 13 changelog announcing the removal of community.mysql in favor of ansible.mysql in Ansible 15 as soon as possible

Filed https://github.com/ansible-community/ansible-build-data/pull/678.

felixfontein: Yes, I'm mainly concerned about runtime deprecation warnings.

gotmax23 (approved):
Maybe the comment should be updated to mention that this is specifically about the _runtime_ deprecation warnings, but +1 otherwise.

13/ansible-13.constraints
mariolenz: OK, you've convinced me. There won't be that many Ansible 13 releases, anyway.

mariolenz (approved):

13/ansible-13.constraints
felixfontein:
```suggestion
# Since ansible.mysql was just added to Ansible, let's wait with this to avoid runtime deprecation warnings.
```

gotmax23 (approved):
Thanks!

felixfontein:
@gotmax23 @mariolenz @Andersson007 thanks!

PR #680, opened 2026-05-19
Pin netbox.netbox due to breaking change in 3.23.0

gotmax23:


mariolenz (approved):
LGTM

mariolenz:
@felixfontein What do you think?

PR #686, opened 2026-05-21
Ansible 14: restrict community.mysql to < 5.0.0

felixfontein:
In case @Andersson007 does not respond to the questions in the community matrix room, we need to either restrict ansibe.mysql to < 5.0.0, or community.mysql to < 5.0.0. I think the latter is better.


gotmax23:
Can you update the `removal:` entries for 13 and 14 to bump the `redirect_replacement_major_version` for community.mysql to `15` if we're not going to include the release that introduces the redirects in Ansible 14?

PR #698, opened 2026-06-02
Add 15/ directory

### LINKED ISSUE #695: Remove cyberark.pas from Ansible 15
See https://forum.ansible.com/t/vote-ended-on-2026-06-01-removal-from-ansible-cyberark-pas/45816/ and https://github.com/ansible-community/ansible-build-data/pull/694/.

mariolenz:
Fixes #645
Fixes #695

Now that 14.0.0 is out, let's create the directory for the next major version.


felixfontein (approved):

mariolenz:
Thx for reviewing @felixfontein!

15/collection-meta.yaml
```diff
@@ -0,0 +1,678 @@
...
+  vultr.cloud:
+    maintainers:
+      - resmo
+      - optik-aper
+    repository: https://github.com/vultr/ansible-collection-vultr
+  vyos.vyos:
+    repository: https://github.com/ansible-collections/vyos.vyos
+  wti.remote:
+    repository: https://github.com/wtinetworkgear/wti-collection
+removed_collections:
```
gotmax23, line 328: Are collections removed in previous releases supposed to be kept here?

mariolenz: I'm not sure if I understand your question. Do you want to know if we should keep collections removed in, in this case, Ansible 14 or earlier under `removed_collections`?

If yes, the answer is I don't know. I would be happy to remove them completely, but I'm not sure if this would break things or not. Maybe @felixfontein knows more.

If no, please explain a bit more what your question is about.

felixfontein: I think we had this discussion already in the past, but I forgot when, where, and what the result was ;) I'm tending to keep the information a bit longer, since this is also used to generate stub pages on the docsite for removed collections - for example, https://docs.ansible.com/projects/ansible/14/collections/dellemc/os10/.

PR #724, opened 2026-08-14
Remove the netapp_eseries.santricity pin

felixfontein:
Ref: https://github.com/NetApp/santricity/issues/36


gotmax23 (approved):
Technically, this is still wrong AFAIU since deprecations aren't supposed to happen in semver minor releases of collections, but I guess this doesn't make much of a difference for the package since we're not in feature freeze.

Thanks for handling this!

felixfontein:
Yes; same as the new feature in 2.0.2...

felixfontein:
@gotmax23 thanks for reviewing!

PR #726, opened 2026-08-16
Harden release workflow: split up build step

### LINKED ISSUE #725: Harden release workflow by splitting build step into build and create PR step
Right now, there's one big build step, that has read-write permissions both to the repository and to PRs in the repository.

It would be better to split this step up into two steps, so that the first step that builds the Ansible package only has minimal rights needed. That step can upload the changed/created/built files, and then a second step can download them and create a PR with them. Only this second step actually needs the full permissions.

felixfontein:
Fixes #725.

### DIFF
--- a/.github/workflows/ansible-release.yml
+++ b/.github/workflows/ansible-release.yml
@@ -210,20 +210,20 @@ jobs:
           ACTOR: "${{ github.actor }}"
         run: |
           body="./pr-body"
-          echo -e "${CI_COMMIT_MESSAGE}\nRelease artifacts: <${ARTIFACT_URL}>\n" >>$body
+          echo -e "${CI_COMMIT_MESSAGE}\nRelease artifacts: <${ARTIFACT_URL}>\n" >> "${body}"
           if [ "${IGNORE_FEATURE_FREEZE}" = "true" ]; then
-            echo -e '- This release was created with `ignore-feature-freeze`.' >>$body
+            echo -e '- This release was created with `ignore-feature-freeze`.' >> "${body}"
           fi
           if [ "${IGNORE_VERSION_REGRESSIONS}" = "true" ]; then
-            echo -e '- This release was created with `ignore-version-regressions`.' >>$body
+            echo -e '- This release was created with `ignore-version-regressions`.' >> "${body}"
           fi
           echo -n "pr-url=" >> "${GITHUB_OUTPUT}"
           gh pr create \
-          --base main \
-          --head "${ANSIBLE_BRANCH_NAME}" \
-          --title "Release Ansible ${ANSIBLE_VERSION}" \
-          --body-file "${body}" \
-          --assignee "${ACTOR}" \
+            --base main \
+            --head "${ANSIBLE_BRANCH_NAME}" \
+            --title "Release Ansible ${ANSIBLE_VERSION}" \
+            --body-file "${body}" \
+            --assignee "${ACTOR}" \
           | tee -a "$GITHUB_OUTPUT"
 
   # publish job downloads the arifacts and publish it to PyPI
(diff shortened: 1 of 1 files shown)


.github/workflows/ansible-release.yml
konstruktoid, line 49: A suggestion is to set `permissions: {}` before any jobs.

felixfontein: Thanks, I also applied that to https://github.com/ansible-community/antsibull-nox/pull/251 and https://github.com/ansible-community/images/pull/163. It's not strictly necessary when Zizmor is used (since that one will complain about every job without explicit permissions from the job or workflow), but it definitely won't hurt even then.

gotmax23 (approved):
Looks good. It'd be nice to test run the workflow before merging it.

felixfontein:
I squashed the commits, replaced the PyPI action with a `run: ls -lah dist/`, pushed that to my fork's `main` branch, enabled GHAs, and triggered the workflow. After fixing an error (see my last commit), it ran through:

* Workflow run: https://github.com/felixfontein/ansible-build-data/actions/runs/33545941700
* Created PR: https://github.com/felixfontein/ansible-build-data/pull/1
* Created tag: https://github.com/felixfontein/ansible-build-data/releases/tag/14.4.0

(Once this PR is merged, I'm going to force-push to my `main` to clean it up, delete the tag, and disable GHAs again.)

.github/workflows/ansible-release.yml
```diff
@@ -146,52 +138,100 @@ jobs:
...
+      pr-url: ${{ steps.create-pr.outputs.pr-url }}
+
+    steps:
+      - name: Pre-create build directory
+        run: mkdir -p antsibull-build/build
+
+      - name: Check out ansible-build-data under antsibull-build build directory
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          # This is where the antsibull-build build-release role expects it by default
```
gotmax23: This comment is not relevant for the create-pr workflow. And I guess we don't need `path:` at all here since we're only working with the a-b-d checkout in the PR step.

.github/workflows/ansible-release.yml
```diff
@@ -252,7 +292,7 @@ jobs:
 
       - name: Create git tag
         env:
-          MERGE_COMMIT: ${{steps.merge-commit.outputs.merge_commit}}
+          MERGE_COMMIT: ${{steps.merge-commit.outputs.merge-commit}}
```
gotmax23: While we're at it, I guess we can change this to use consistent formatting too :)?


```suggestion
          MERGE_COMMIT: ${{ steps.merge-commit.outputs.merge-commit }}
```

felixfontein: 3cf596842c81d2c5c6b780483daea41f9425c293

.github/workflows/ansible-release.yml
felixfontein: c7908ce928412c87bfd3a9082fc751ab5495f578 and 45a53ebce613df5bf9262982d1b57ba89caf0d05.

gotmax23:
Thanks for working on this! I'll try to look at this again during $DAYTIME_HOURS, but I think we're in good shape to get this merged before next week's release unless there's any further comments from other release managers.

.github/workflows/ansible-release.yml
gotmax23, line 91: Future enhancement: I wonder if we should consider pinning these packages and their dependencies, but that makes things difficult since we always want the latest `antsibull-build` and `ansible-core`.

gotmax23 (approved):
Okay, I gave this another look-through. LGTM!

.github/workflows/ansible-release.yml
felixfontein: The latest ansible-core isn't *that* important here, but the latest antsibull-build is. Having some pinning workflow that can be triggered when needed would probably be fine, but that's definitely something for the future...

felixfontein:
(I removed the traces of the workflow run from my repo, in case anyone wonders.)

gotmax23:
There has been no further feedback, so let's get this merged for next week's release. Thanks Felix!

felixfontein:
@konstruktoid @gotmax23 thanks a lot for reviewing and merging!

